PR #4, opened 2024-09-29
Rename `Build` step to `Build / Test`

ahoppen:
The `Build` step usually also runs tests. Rename the step’s name to reflect that.

### DIFF
--- a/.github/workflows/swift_package_test.yml
+++ b/.github/workflows/swift_package_test.yml
@@ -58,5 +58,5 @@ jobs:
           done
     - name: Pre-build
       run: ${{ inputs.pre_build_command }}
-    - name: Build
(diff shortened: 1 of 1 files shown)


.github/workflows/swift_package_test.yml
```diff
@@ -58,5 +58,5 @@ jobs:
           done
     - name: Pre-build
       run: ${{ inputs.pre_build_command }}
-    - name: Build
+    - name: Build / Test
```
MahdiBM, line 61: `/` is what GitHub itself uses in names like `test / api-breakage (pull_request)`.
Should we avoid using `/` to not ambiguate where that slash is coming from?

shahmishal (approved):

PR #12, opened 2024-10-01
Fix the version of swift-format being used in format check

ahoppen:
I think we should pin down one specific version of swift-format to be used in the format check to avoid CI failures if swift-format changes.

### DIFF
--- a/.github/workflows/soundness.yml
+++ b/.github/workflows/soundness.yml
@@ -143,7 +143,7 @@ jobs:
     if: ${{ inputs.format_check_enabled }}
     runs-on: ubuntu-latest
     container:
-      image: swiftlang/swift:nightly-6.0-jammy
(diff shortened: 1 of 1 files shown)


.github/workflows/soundness.yml
```diff
@@ -143,7 +143,7 @@ jobs:
     if: ${{ inputs.format_check_enabled }}
     runs-on: ubuntu-latest
     container:
-      image: swiftlang/swift:nightly-6.0-jammy
+      image: swift:6.0.1-jammy
```
shahmishal, line 146:
```suggestion
      image: swift:6.0-jammy
```

This will allow us to pick up dot releases without needing new PR.

ahoppen: Updated

shahmishal (approved):

PR #14, opened 2024-10-03
Add a soundness check to lint Python files

ahoppen:


.github/workflows/soundness.yml
```diff
@@ -177,3 +185,21 @@ jobs:
...
+    runs-on: ubuntu-latest
+    container:
+      image: ${{ inputs.python_lint_container_image }}
+    timeout-minutes: 5
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+      - name: Run shellcheck
```
shahmishal, line 200:
```suggestion
      - name: Run Python lint
```

.github/workflows/soundness.yml
```diff
@@ -177,3 +185,21 @@ jobs:
         run: |
           apt-get -qq update && apt-get -qq -y install shellcheck
           git ls-files -z '*.sh' | xargs -0 shellcheck
+  python-lint:
+    name: Python lint
+    if: ${{ inputs.python_lint_enabled }}
+    runs-on: ubuntu-latest
+    container:
+      image: ${{ inputs.python_lint_container_image }}
```
FranzBusch, line 193: Similar comment as in your other PR I don't think we need a container here. The base runner comes with python and pip installed.

ahoppen: Nice. Good suggestion. I updated it.

shahmishal (approved):
LGTM!

shahmishal:
@FranzBusch, please feel free to merge it once you’ve reviewed it.

FranzBusch (approved):

FranzBusch:
I just merged your other PR @ahoppen but now that I think about it again we might want to fold this into the Swift format PR after all. The reason why is that each new check here is producing a separate check on the repos. Now most repos don't have all the languages. It might be better to just have one general formatting job that does
- Shellcheck
- swift-format
- yamlint
- python lint
- etc.
WDYT?

ahoppen:
I don’t have a strong personal opinion here: 

I think the advantages of having a single job are:
- We only create a single item in the status check for a PR, which is less noisy

The disadvantages are:
- Running `swift-format` requires us to use a Swift container, all other tools can run on the image itself. So, either we need to install `yamllint` etc. in the container or have one job for `swift-format` and one for all the things that can run on the runner itself.
- If eg. `yamllint` fails, we probably won’t continue execution, which means that you might only find the `shellcheck` issues after fixing all YAML issues. Probably not a huge problem in practice though.
- If you disable a check, we would just skip it in the job and wouldn’t mark it as skipped in the GitHub PR check UI. Not sure if that’s an advantage or disadvantage though.

@shahmishal What’s your opinion?

shahmishal:
Having checks for each lang format/lint makes it easier to figure out what failed. Normally people will ignore all of the green check marks and focus on failures.

ahoppen:
OK, sounds like we want to stick with separate jobs for each lint then.

.github/workflows/soundness.yml
```diff
@@ -200,3 +204,18 @@ jobs:
         run: |
           curl -s https://raw.githubusercontent.com/swiftlang/github-workflows/refs/heads/main/.github/workflows/configs/yamllint.yml > /tmp/yamllint.yml
           yamllint --strict --config-file /tmp/yamllint.yml ${GITHUB_WORKSPACE}
+  python-lint:
```
FranzBusch, line 207:
```suggestion
  python-lint-check:
```

.github/workflows/soundness.yml
```diff
@@ -200,3 +204,18 @@ jobs:
...
+    name: Python lint
```
FranzBusch, line 208:
```suggestion
    name: Python lint check
```

FranzBusch (commented):
Okay let's proceed with this then. Just some naming nits to stay consistent

ahoppen:
Should we go the opposite way and remove `check` from all of these jobs? It feels redundant to me because I would expect all CI jobs to check something.

ahoppen:
Discussed offline that we want to use `check` in the name.

.github/workflows/soundness.yml
```diff
@@ -200,3 +204,18 @@ jobs:
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: false
+      - name: Run flake8
+        run: |
+          pip3 install flake8 flake8-import-order --break-system-packages
```
kkebo, line 219: As you may know, it seems that `ubuntu-latest` on GitHub-hosted runners is currently in the process of migrating to Ubuntu 24.04. Until it's completed, Ubuntu 22.04 may be used, and on Ubuntu 22.04, this command will fail because its `pip3` is old and does not have `--break-system-packages`.

Announcement: https://github.com/actions/runner-images/issues/10636

Failure case: https://github.com/kkebo/swift-format/actions/runs/11299915833/job/31431847497

Since the migration will be completed soon, I think we can ignore it.

ahoppen: Just fixed it: https://github.com/swiftlang/github-workflows/pull/28

kkebo: Thank you!

PR #16, opened 2024-10-08
Support gradle and groovy files in check-license-header.sh

ktoso:
.gradle and .groovy files are common in build configuration when building Java projects with Gradle, therefore we should support them as well.


shahmishal (approved):

ktoso:
Thank you!

PR #17, opened 2024-10-09
Add new job : check local swift dependencies

GErP83:


FranzBusch:
Could you explain what the intention behind this check is?

GErP83:
Sometimes in Swift development, we use local dependencies, such as:
```
dependencies: [
        .package(path: "../SomePackage")
    ],
```
However, we might forget to remove or replace them later. This script can help by providing a warning in such cases.

.github/workflows/soundness.yml
```diff
@@ -51,6 +51,10 @@ on:
         type: string
         description: "Container image for the shell check job. Defaults to latest Swift Ubuntu image."
         default: "swift:6.0-noble"
+      local_swift_dependencies_check_enabled:
+        type: boolean
+        description: "Boolean to enable the local swift dependencies check job. Defaults to true."
+        default: true
```
ahoppen, line 57: All the Swift packages that are built into the toolchain will have local Swift dependencies, eg. https://github.com/swiftlang/swift-format/blob/d143a8a59c89aba5bdadf96787f7f41197ce6b71/Package.swift#L173-L175. I’m thus not sure if we want to enable this by default.

tib: I agree, this should be disabled by default. Users should be able to opt-in for this check. Apart from that I think this is very useful in some cases. :)

GErP83: Changed the the default value to **false**.

FranzBusch (changes requested):
As @ahoppen pointed out there are use-cases where local dependencies are expected such as packages used in the toolchain. Additionally, a lot of repositories are using nested packages that reference the root package for benchmarks, examples and much more. 

At this point, I don't see a real need to enforce no local package dependencies across our repositories and would like us to wait with this until we see it becoming a problem.

PR #26, opened 2024-10-11
Align some the YAML lint namings in the soundness workflow

FranzBusch:


ahoppen (commented):
Should we go the opposite direction and remove `check` from the jobs? I mean, what else would CI jobs do other than checking that something passes?

FranzBusch:
> Should we go the opposite direction and remove `check` from the jobs? I mean, what else would CI jobs do other than checking that something passes?

I am reluctant to any unnecessary naming changes since it would mean we have to update branch protection rules in all the repos that have adopted this already

ahoppen (approved):

PR #28, opened 2024-10-12
Remove `--break-system-packages` from Python lint job

ahoppen:
`--break-system-packages` was only needed when we ran in a Swift container and is an unknown option when running directly on the GitHub action runner.

Also change the pull request workflow on this repo to use the workflows they are modified by PRs, not as they are defined on `main` before the PR was applied. This would have caught the Python-lint issue and allows me to unbreak it.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -7,7 +7,7 @@ on:
 jobs:
   soundness:
     name: Soundness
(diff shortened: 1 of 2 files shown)


ahoppen:
Python lint is failing with the issue I’m fixing in this PR.

shahmishal (approved):

PR #29, opened 2024-10-14
Ignore some python formatting errors for licenses

FranzBusch:
# Motivation

We want to lint our python files but also need to include our license header.

# Modification

This PR ignores a few python flake rules to make sure the license headers pass.

# Result

Passing license headers.

### DIFF
--- a/.github/workflows/soundness.yml
+++ b/.github/workflows/soundness.yml
@@ -218,4 +218,8 @@ jobs:
         run: |
           pip3 install flake8 flake8-import-order
           cd ${GITHUB_WORKSPACE}
+          if [ ! -f ".flake8" ]; then
+            echo "Downloading default flake8 config file"
+            curl -s https://raw.githubusercontent.com/swiftlang/github-workflows/refs/heads/main/.github/workflows/configs/.flake8.yml > .flake8
+          fi
           flake8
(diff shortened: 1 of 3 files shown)


.github/workflows/soundness.yml
```diff
@@ -218,4 +218,4 @@ jobs:
         run: |
           pip3 install flake8 flake8-import-order
           cd ${GITHUB_WORKSPACE}
-          flake8
+          flake8 --extend-ignore=E265,E266,E501
```
ahoppen, line 221: Completely disabling maximum line length (E501) seems a little to lenient to me. I think projects should be able to enforce that.

What do you think about the following instead?
- Instead of passing an ignore list to `flake8` here, we install a default `.flake8` file to `/` on the runner. That way we use the default `.flake8` unless a repo overrides it.
- In that default format file, we set `max-line-length = 88`, matching what the [compiler repo does](https://github.com/swiftlang/swift/blob/3aed0950aaeffca58c283f703b2be3d1dafd10bd/.flake8#L68). This would also fit license header

FranzBusch:
The CI failure here is expected since the default conflict is being pulled from main.

ahoppen (approved):

PR #36, opened 2024-10-17
allow projects to use a custom license header template

sebsto:


ktoso (approved):
LGTM, this is necessary for some projects like the lambda runtime which must retain e.g. Amazon license header because of @sebsto being primary contributor etc.

PR #38, opened 2024-10-17
Fix license header check ignoring well known files

FranzBusch:
The previous PR tried to ignore two well known files `.licenseignore` and `.license_header_template`. However, that PR wasn't fully working and broke this check.

This PR fixes it by using a slightly different approach for those well known files.

### DIFF
--- a/.github/workflows/scripts/check-license-header.sh
+++ b/.github/workflows/scripts/check-license-header.sh
@@ -1,15 +1,15 @@
 #!/bin/bash
-# ===----------------------------------------------------------------------===//
-#
-# This source file is part of the Swift.org open source project
-#
-# Copyright (c) 2024 Apple Inc. and the Swift project authors
-# Licensed under Apache License v2.0 with Runtime Library Exception
-#
-# See https://swift.org/LICENSE.txt for license information
-# See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
-#
-# ===----------------------------------------------------------------------===//
+## ===----------------------------------------------------------------------===##
+##
+## This source file is part of the Swift.org open source project
+##
+## Copyright (c) 2024 Apple Inc. and the Swift project authors
+## Licensed under Apache License v2.0 with Runtime Library Exception
+##
+## See https://swift.org/LICENSE.txt for license information
+## See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+##
+## ===----------------------------------------------------------------------===##
 
 set -euo pipefail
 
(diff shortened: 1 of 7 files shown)


.licenseignore
```diff
@@ -0,0 +1,5 @@
+.github/workflows/configs/.flake8
+**/*.yml
+CODEOWNERS
+LICENSE.txt
+README.md
```
ahoppen, line 5: Should these files have trailing newlines?

FranzBusch:
We can also enable the license check on this repo after this change to make sure it stays green

.license_header_template
```diff
@@ -0,0 +1,11 @@
...
+#
+# This source file is part of the Swift.org open source project
+#
+# Copyright (c) YEARS Apple Inc. and the Swift project authors
+# Licensed under Apache License v2.0 with Runtime Library Exception
+#
+# See https://swift.org/LICENSE.txt for license information
+# See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+#
+# ===----------------------------------------------------------------------===#
```
ahoppen, line 11: What do we need the license header template for? It wasn’t there before

.github/workflows/scripts/check-license-header.sh
```diff
@@ -40,12 +40,12 @@ fi
...
 
-file_excludes=".license_header_template
-.licenseignore"
-if [ -f .licenseignore ]; then 
-  file_excludes=$file_excludes$(printf '\n')$(cat .licenseignore)
+if [[ -f .licenseignore ]]; then
+  file_paths=$(tr '\n' '\0' < .licenseignore | xargs -0 -I% printf '":(exclude)%" '| xargs git ls-files ":(exclude).licenseignore" ":(exclude).license_header_template" )
+else
+  file_paths=$(git ls-files ":(exclude).license_header_template" )
 fi
```
ahoppen, line 47: The second `xargs` confused me a little bit until I realized that it only causes `git ls-files` to be called a single time.  Maybe worth pulling out the computation of the exclude list to make it more readable? Something along the lines of the following.
 
```suggestion
if [[ -f .licenseignore ]]; then
  exclude_list="':(exclude).licenseignore' ':(exclude).license_header_template' $($(tr '\n' '\0' < .licenseignore | xargs -0 -I% printf '":(exclude)%")"
else
  exclude_list=":(exclude).license_header_template"
fi
file_paths=$(git ls-files $exclude_list)
```

FranzBusch: I agree though I would prefer a follow up PR to unblock the CI on other repos

ahoppen (approved):
Accepting to unblock CI and let’s address https://github.com/swiftlang/github-workflows/pull/38/files#r1805237145 in a follow-up PR.

Issue #34, opened 2024-10-16
Add option to build with the Static Linux SDK


0xTim:
Currently the workflows provided have the ability to build a number of versions of Swift. It would be great to add a workflow to be able to test compilation with the Static Linux SDK. We do something similar [in Vapor](https://github.com/vapor/ci/blob/main/.github/workflows/run-unit-tests.yml#L271-L282) - happy to provide a PR if that would be useful

shahmishal:
This is good idea, we should include Static Linux SDK testing. 

>happy to provide a PR if that would be useful

Please! Thank you

PR #40, opened 2024-10-21
Add workflow for testing Musl

0xTim:
Resloves #34 

Need to decide how to handle different versions and probably improve the workflow for installing the correct static SDK but opening as is for now for discussion


.github/workflows/swift_package_test.yml
```diff
@@ -27,6 +27,10 @@ on:
         type: string
         description: "Linux command to execute before building the Swift package"
         default: ""
+      musl_pre_build_command:
+        type: string
+        description: "Linux command to execute before building the Swift package"
```
justice-adams-apple, line 32: I think the description here just needs updated, as it's the same as the linux_pre_build_command

.github/workflows/swift_package_test.yml
```diff
@@ -110,3 +122,27 @@ jobs:
...
+    if: ${{ inputs.enable_musl_checks }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    container: swift:6.0-noble
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v4
+      - name: Install SDK
+        run: swift sdk install https://download.swift.org/swift-6.0.1-release/static-sdk/swift-6.0.1-RELEASE/swift-6.0.1-RELEASE_static-linux-0.0.1.artifactbundle.tar.gz --checksum d4f46ba40e11e697387468e18987ee622908bc350310d8af54eb5e17c2ff5481
```
justice-adams-apple, line 137: do we want a hardcoded checksum/artifact url hardcoded here? Or should they be inputs

.github/workflows/swift_package_test.yml
```diff
@@ -110,3 +122,27 @@ jobs:
...
         timeout-minutes: 60
         run: docker run -v ${{ github.workspace }}:C:\source swiftlang/swift:${{ matrix.swift_version }}-windowsservercore-1809 cmd /s /c  "swift --version & swift test --version & cd C:\source\ & ${{ inputs.windows_build_command }} ${{ inputs.swift_nightly_flags }}"
+
+  musl-build:
+    name: MUSL
+    if: ${{ inputs.enable_musl_checks }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    container: swift:6.0-noble
```
sebsto, line 132: The version here must be exactly the same as the SDK below.  As of today (30/10) `swift:6.0-noble` vends Swift 6.0.2 which is not working with the SDK 6.0.1.

I suggest to pass both the Swift version number and the checksum as parameters

0xTim:
The NIO team are working on something that actually works - https://github.com/apple/swift-nio/pull/3019

shahmishal:
Supported add it recently - https://github.com/swiftlang/github-workflows/pull/142

shahmishal:
Please re-open the PR if it does not solve the problem you were facing.

PR #45, opened 2024-10-24
Allow execution of a pre-build command before testing a package on Windows

ahoppen:
The motivating change for this is https://github.com/swiftlang/swift-format/pull/863.

While at it, also unify the release versions and nightly versions of Windows into a single matrix, avoiding code duplications and also making them show up in a single matrix in the GitHub workflow run, which looks nicer.

Example run: https://github.com/ahoppen/swift-format/actions/runs/11502681752

### DIFF
--- a/.github/workflows/swift_package_test.yml
+++ b/.github/workflows/swift_package_test.yml
@@ -29,18 +29,22 @@ on:
         default: ""
       linux_build_command:
         type: string
-        description: "Linux Build command default is swift test"
+        description: "Linux command to build and test the package"
         default: "swift test"
+      windows_pre_build_command:
+        type: string
+        description: "Windows Command Prompt command to execute before building the Swift package"
+        default: ""
       windows_build_command:
         type: string
-        description: "Linux Build command default is swift test"
+        description: "Windows Command Prompt command to build and test the package"
         default: "swift test"
       linux_env_vars:
         description: "List of environment variables"
         type: string
       enable_windows_checks:
         type: boolean
-        description: "Boolean to enable windows testing. Defaults to true."
+        description: "Boolean to enable windows testing. Defaults to true"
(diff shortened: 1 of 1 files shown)


.github/workflows/swift_package_test.yml
```diff
@@ -76,37 +80,44 @@ jobs:
...
-        run: docker pull swiftlang/swift:${{ matrix.swift_version }}-windowsservercore-1809
-      - name: Checkout repository
-        uses: actions/checkout@v4
+        run: |
+          if ("${{ matrix.swift_version }}".Contains("nightly")) {
+            $Image = "swiftlang/swift:${{ matrix.swift_version }}-windowsservercore-1809"
+          } else {
+            $Image = "swift:${{ matrix.swift_version }}-windowsservercore-ltsc2022"
+          }
+          docker pull $Image
```
justice-adams-apple, line 100: could we have save the image name as output just to remove duplication below:
```suggestion
          docker pull $Image
          echo "image=$Image" >> "$GITHUB_OUTPUT"
```

Then in the "Create test script" section we can do
```yaml
docker run -v ${{ github.workspace }}:C:\source -v $env:TEMP\test-script:C:\test-script ${{ steps.pull-docker-image.outputs.image }}" C:\test-script\run.cmd
```

(need to set the id of the step to `pull-docker-image`)

See: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-an-output-parameter

ahoppen: Great suggestion. Updated and kicked off another test run: https://github.com/ahoppen/swift-format/actions/runs/11505380578

justice-adams-apple (approved):

PR #47, opened 2024-10-24
Use Powershell to run tests on Windows

ahoppen:
This allows us to run tests on Windows with Swift 5.10.

Test run with relevant configurations: https://github.com/ahoppen/swift-format/actions/runs/11508000825

### DIFF
--- a/.github/workflows/swift_package_test.yml
+++ b/.github/workflows/swift_package_test.yml
@@ -84,7 +84,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        swift_version: ['5.9', '6.0', 'nightly', 'nightly-6.0']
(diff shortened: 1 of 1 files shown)


.github/workflows/swift_package_test.yml
```diff
@@ -84,7 +84,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        swift_version: ['5.9', '6.0', 'nightly', 'nightly-6.0']
+        swift_version: ['5.9', '5.10', '6.0', 'nightly']
```
shahmishal, line 87: Is there a reason we are dropping `nightly-6.0`?

ahoppen: No, just forgot to add it back in after I removed it so testing uses fewer of my free GitHub Action minutes.

shahmishal (approved):

PR #52, opened 2024-10-29
Support cross-PR testing for Swift packages

ahoppen:
This allows Swift Packages to be tested in combination with PRs updating their dependencies. To reference a linked PR, `Linked PR: <link to PR>` needs to be added to the PR description, eg:

```
Linked PR: https://github.com/swiftlang/swift-syntax/pull/2859
```

Example run: https://github.com/swiftlang/swift-format/actions/runs/11564842847/job/32193164363

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -68,4 +68,3 @@ jobs:
     with:
       api_breakage_check_enabled: false
       license_header_check_project_name: "Swift.org"
-      format_check_enabled: false
--- a/.github/workflows/scripts/cross-pr-checkout.swift
+++ b/.github/workflows/scripts/cross-pr-checkout.swift
@@ -1,0 +1,247 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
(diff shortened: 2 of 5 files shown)


FranzBusch (commented):
Looks good in general to me. What is the expectation post merge of a PR that has linked PRs. Couldn't it be that the linked PRs haven't been merged or that deps haven't been updated and main becomes unstable?

ahoppen:
Yes, you will need to merge your dependency PRs before the PR itself. Otherwise the post-merge check will fail but will recover once the dependencies have been updated. I don’t really think there is any way around that.

FranzBusch:
Instead of doing this in the regular checks that are triggered on the PR events we could make this triggered by leaving a comment on the PR. 
This way you can trigger it manually and get feedback if it works. Then merge the PRs in your deps. Once they are merged you can run the normal PR checks.
Would that work?

ahoppen:
> Instead of doing this in the regular checks that are triggered on the PR events we could make this triggered by leaving a comment on the PR.
This way you can trigger it manually and get feedback if it works. Then merge the PRs in your deps. Once they are merged you can run the normal PR checks.
Would that work?

No, that doesn’t work. https://github.com/swiftlang/swift-syntax/pull/2859, for example, introduces a breaking swift-syntax change that needs a corresponding update in swift-format, since swift-format is also in the toolchain (in this case, that changes is done by https://github.com/swiftlang/swift-format/pull/866). Thus, we need to be able to get both the swift-syntax and the swift-format PR into a green CI state at the same time so that the *Merge pull request* button can be hit on both of them at (roughly) the same time. This is how cross-PR testing works in then Jenkins-based Swift CI for toolchain repos at the moment (https://github.com/swiftlang/swift/blob/main/docs/ContinuousIntegration.md#cross-repository-testing).

PR #55, opened 2024-11-02
Add a workflow for welcoming first-time contributors

AnthonyLatsis:
I tested the workflow in my [swift fork](https://github.com/AnthonyLatsis/swift/actions), albeit without the first-time contributor condition. Once this lands, we can use the contributor experience repo to further test it against pull requests from forks.


.github/workflows/comment_on_first_contribution.yml
```diff
@@ -0,0 +1,56 @@
...
+  comment:
+    if: >
+      (
+        (github.event_name == 'pull_request_target' &&
+         inputs.pull_request_message) ||
+        (github.event_name == 'issues' && inputs.issue_message)
+      ) &&
+      github.event.action == 'opened' &&
+      contains(
+        fromJSON('["FIRST_TIMER", "FIRST_TIME_CONTRIBUTOR"]'),
```
ahoppen, line 29: Is `fromJSON` needed here or could this just be the following?

```suggestion
        ["FIRST_TIMER", "FIRST_TIME_CONTRIBUTOR"],
```

.github/workflows/comment_on_first_contribution.yml
```diff
@@ -0,0 +1,56 @@
...
+        github.event_name == 'issues'
+          && github.event.issue.author_association
+          || github.event.pull_request.author_association
+      )
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
```
ahoppen, line 37: I don’t think this is necessary. According to the GitHub documentation, workflows from the `pull_request_target` already have write access.

.github/workflows/comment_on_first_contribution.yml
```diff
@@ -0,0 +1,56 @@
...
+        required: false
+
+jobs:
+  comment:
+    if: >
+      (
+        (github.event_name == 'pull_request_target' &&
+         inputs.pull_request_message) ||
+        (github.event_name == 'issues' && inputs.issue_message)
+      ) &&
```
ahoppen, line 26: I think this would read a lot easier if you remove one line break. We don’t enforce a maximum line length on YAML.

```suggestion
      (
        (github.event_name == 'pull_request_target' && inputs.pull_request_message) ||
        (github.event_name == 'issues' && inputs.issue_message)
      ) &&
```

README.md
```diff
@@ -73,6 +73,29 @@ pre_build_command: "apt-get update -y -q && apt-get install -y -q example"
 
 macOS and Windows platform support will be available soon.
 
+### Welcome a first-time contributor
+
+This is example of a recommended workflow for creating a comment with the
+specified Markdown text when a first-time contributor opens an issue or pull
+request:
```
ahoppen, line 80: I think it’s worth mentioning that if `pull_request_message` or `issue_message` is omitted, no comment will be posted on PRs / issues.

ahoppen (commented):
I don’t think this workflow design will work because GitHub Actions will be disabled for first-time contributors and thus the action will not trigger on the `pull_request_target

.github/workflows/comment_on_first_contribution.yml
AnthonyLatsis: But this can be triggered by issues too.

.github/workflows/comment_on_first_contribution.yml
AnthonyLatsis: There appears to be no better way to [construct an array](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#example-matching-an-array-of-strings). No array literals either.

AnthonyLatsis:
> I don’t think this workflow design will work because GitHub Actions will be disabled for first-time contributors and thus the action will not trigger on the `pull_request_target

I am reading through [this](https://docs.github.com/en/organizations/managing-organization-settings/disabling-or-limiting-github-actions-for-your-organization#configuring-required-approval-for-workflows-from-public-forks) and getting the impression that the limitation applies to workflow runs — regardless of whether they depend on actions. But the note says that `pull_request_target` is never run in a (unsafe) fork context and, thus, is an exception. I assume the `issues` event could similarly be triggered by anyone because the workflow would always run in the context of the repository.

.github/workflows/comment_on_first_contribution.yml
AnthonyLatsis: Oh, maybe you meant that we can rid of the `pull-requests` key. That makes sense 👍

ahoppen:
> I am reading through [this](https://docs.github.com/en/organizations/managing-organization-settings/disabling-or-limiting-github-actions-for-your-organization#configuring-required-approval-for-workflows-from-public-forks) and getting the impression that the limitation applies to workflow runs — regardless of whether they depend on actions. But the note says that pull_request_target is never run in a (unsafe) fork context and, thus, is an exception. I assume the issues event could similarly be triggered by anyone because the workflow would always run in the context of the issue’s repository.

I’m sorry. I meant to remove that comment after I realized that you’re using `pull_request_target` but must have forgotten.

.github/workflows/comment_on_first_contribution.yml
ahoppen: Oh, wow. I didn’t realize that. Fair enough then.

ahoppen (approved):
Looks good to me but I would like @shahmishal to review this as well because it’s a new workflow.

AnthonyLatsis:
Sure, I’d like that too! Thanks for reviewing, Alex.

.github/workflows/comment_on_first_contribution.yml
```diff
@@ -0,0 +1,54 @@
...
+          The Markdown message to comment when a contributor creates their first
+          pull request.
+        type: string
+        required: false
+      issue_message:
+        description: >
+          The Markdown message to comment when a contributor creates their first
+          issue.
+        type: string
+        required: false
```
FranzBusch, line 17: Can we provide a default message for both? Would avoid every repo to define its own custom message. Maybe even going one step further do we want each repo to customise this really? If so how do you expect them to customise this?

PR #58, opened 2024-11-05
Include changes to the target branch when action is re-run

ahoppen:
When a GitHub action is re-run, it uses the same context as the original run. This means that it runs on the exact same merge commit and does not pick up changes done to the base branch since the last run.

In case the base branch is was failing when a PR was opened, this means that we can’t get CI to pass by fixing the base branch without pushing a new commit to the PR branch.

Adding `ref: ${{ github.ref }}` to the `actions/checkout` action works around this by checking out the PR merge ref (ie. `refs/pull/<pr_number>/merge`) instead of a fixed commit and the PR merge ref is updated, as described in https://github.com/actions/checkout/issues/1036.

### DIFF
--- a/.github/workflows/soundness.yml
+++ b/.github/workflows/soundness.yml
@@ -83,6 +83,7 @@ jobs:
         with:
           persist-credentials: false
           submodules: true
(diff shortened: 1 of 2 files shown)


ktoso (approved):
Sounds good to me 👍

shahmishal (approved):

PR #60, opened 2024-11-07
Revert "Include changes to the target branch when action is re-run"

simonjbeaumont:
Reverts swiftlang/github-workflows#58

This has resulted in a destructive outcome for developers using `act` to run the workflows locally, which is also in our public documentation now, particularly when used with `--bind`.

This patch recreates the entire repo and force updates to a ref. It literally _deleted_ my Git repo and left me with a grafted shallow clone. It discarded all my local branches and no reflog was available.

We need to revert this one quickly.

I've verified that the revert mitigates the problem by making this change in my YAMLs locally, to the commit before this PR was merged: 

```diff
-    uses: apple/swift-nio/.github/workflows/cxx_interop.yml@main
+    uses: swiftlang/github-workflows/.github/workflows/soundness.yml@814f5831ccc6aac159903e8db6daba726c8b62d5
```

I'm sympathetic to the motivation behind the PR but I think we should revert immediately to avoid others hitting this.

### DIFF
--- a/.github/workflows/soundness.yml
+++ b/.github/workflows/soundness.yml
@@ -83,7 +83,6 @@ jobs:
         with:
           persist-credentials: false
           submodules: true
-          ref: ${{ github.ref }}  # Include changes to the target branch when action is re-run https://github.com/actions/checkout/issues/1036
       - name: Mark the workspace as safe
         # https://github.com/actions/checkout/issues/766
         run: git config --global --add safe.directory ${GITHUB_WORKSPACE}
@@ -105,7 +104,6 @@ jobs:
         with:
           persist-credentials: false
           submodules: true
-          ref: ${{ github.ref }}  # Include changes to the target branch when action is re-run https://github.com/actions/checkout/issues/1036
       - name: Run documentation check
         run: |
           apt-get -qq update && apt-get -qq -y install curl yq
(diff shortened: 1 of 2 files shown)


simonjbeaumont:
//cc @shahmishal @ahoppen @ktoso @rnro

shahmishal (approved):

simonjbeaumont:
Thanks for merging promptly.

I'm finished for today now but I think we could investigate using a conditional here on the `CI` environment variable and only have this behaviour if running in CI. I believe we have a similar behaviour somewhere else already but I'm on a mobile device so can't validate right now.

ahoppen:
Sorry for breaking the `act` workflow @simonjbeaumont and thanks for reverting my change. 

If you have an example of how you guard behavior on the CI environment, I’d be interested to see that and check if we can apply it here as well.

simonjbeaumont:
@ahoppen Looks like we can use `job.if` here with an environment variable that's only set when using `act`:

```yaml
jobs:
  foo:
    runs-on: ubuntu-latest
    name: Foo
    steps:
      - name: Dump env
        run: env | sort
      - name: Echo always
        run: echo "always"
      - name: Echo only if FOO=true
        if: env.FOO == 'true'
        run: echo "job running with FOO set to true"
      - name: Echo only if not FOO=true
        if: env.FOO != 'true'
        run: echo "job not running with FOO set to true"
      - name: Echo only if running with act
        if: env.ACT == 'true'
        run: echo "job running with ACT"
      - name: Echo only if not running with act
        if: env.ACT != 'true'
        run: echo "job not running with act"
```

So I think it will just be adding `if: env.ACT != 'true'` to the steps. I'll open a PR to reintroduce the change.

PR #66, opened 2024-11-19
Add linux_pre_run_command to soundness

rnro:
Add a `linux_pre_run_command` input parameter to the soundness workflow jobs to allow callers to set up any dependencies or perform any other preparations before the soundness checks are executed.

This is useful for example for repositories which require dependencies to be installed before the build.

The `if:` statement evaluates to false if the input is not defined or is set to `""` so the step does not appear in output.

### DIFF
--- a/.github/workflows/soundness.yml
+++ b/.github/workflows/soundness.yml
@@ -63,6 +63,10 @@ on:
         type: boolean
         description: "Boolean to enable the Python lint check job. Defaults to true."
         default: true
+      linux_pre_run_command:
+        type: string
+        description: "Linux command to execute before building the Swift package"
+        default: ""
 
 ## We are cancelling previously triggered workflow runs
 concurrency:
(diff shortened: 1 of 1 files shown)


.github/workflows/soundness.yml
```diff
@@ -63,6 +63,10 @@ on:
         type: boolean
         description: "Boolean to enable the Python lint check job. Defaults to true."
         default: true
+      linux_pre_run_command:
```
FranzBusch, line 66: Should this really be executed before every soundness check. It seems like this is only relevant for the ones that need to build something e.g. docs and api break

rnro: I wondered that myself but I think it's plausible that a caller may want to do some other source preparations, munging, generation that this would permit.

rnro: I'm not set on it being this way though if you'd rather we solve the immediate requirement.

ktoso: swift-java needs a bunch of weird stuff, like installing a JDK for builds to work.

I agree that it is only for things which "build" the formatting check for example does no need it. At least so far AFAICS

rnro: Yep, I'm happy taking this approach. Thanks for the feedback everyone.

FranzBusch (approved):

PR #74, opened 2024-12-09
Setup a simple test build to verify

### LINKED ISSUE #73: Should have PR checks to make sure builds don't break
We had to quickly patch the shared workflow file with #72, after the Docker workflow broke on Windows. We should setup a simple test step for new pull requests that verifies that PRs that change the workflow file do not break builds downstream.

Should be able to use a sub workflow so changes to the workflow are tested: https://github.com/actions/runner/discussions/1419

award999:
* Will test Linux and Windows docker workflows
* Will test Windows docker-less workflow
* Tests will init and build a tiny swift library
* The "uses" is relative to use the one that may be changed by the PR

Fixes #73

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -5,6 +5,38 @@ on:
     types: [opened, reopened, synchronize]
 
 jobs:
+  tests_with_docker:
+    name: Test with Docker
+    uses: ./.github/workflows/swift_package_test.yml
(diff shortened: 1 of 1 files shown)


ahoppen (approved):
Thank you 🤩

shahmishal:
Thank you!

shahmishal (approved):

PR #80, opened 2024-12-20
Always use windows server 2022 for dockerless workflow

award999:
* LLDB is not working on the 2019 images for the vscode-swift tests

Issue: #1263


award999:
@shahmishal confirmed that's the issue, works downstream https://github.com/swiftlang/vscode-swift/pull/1291

shahmishal (approved):

PR #84, opened 2025-01-08
Revert "Revert "Always use windows server 2022 for dockerless workflow""

award999:
Reverts swiftlang/github-workflows#81


award999:
Talked to @bnbarham and @ahoppen and they'll modify swift-format to work on Windows server 2022

ahoppen (approved):

PR #86, opened 2025-02-01
Update README to reference the correct options

dimitribouniol:
I noticed that the option I copied from the README was no longer functional, so updating the readme accordingly. I also added references to Windows since it seems to be supported now.


shahmishal (approved):

dimitribouniol:
@shahmishal Just wanted to say thanks for the quick reviews on both #85 and #86!

PR #89, opened 2025-02-05
Allow using GITHUB_TOKEN downstream

award999:
Looking at https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context the `github.token` is only available withins the `steps`. Of the common workflow, but the jobs extending it do not have access. The vscode-swift extension publishes a build of the extension in a separate job, and we want to be able to download it within the common workflow so we can test what is actually might be released.

### DIFF
--- a/.github/workflows/swift_package_test.yml
+++ b/.github/workflows/swift_package_test.yml
@@ -60,6 +60,10 @@ on:
         type: boolean
         description: "Boolean to enable running build in windows docker container. Defaults to true"
         default: true
+      needs_token:
+        type: boolean
+        description: "Boolean to enable providing the GITHUB_TOKEN to downstream job."
+        default: false
 
 jobs:
   linux-build:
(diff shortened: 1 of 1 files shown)


award999:
@shahmishal I've tested this on my fork of vscode-swift https://github.com/award999/vscode-swift/actions/runs/13144430476. It does allow us to download the artifact for the sake of testing the extension we may ship.

https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#permissions-for-the-github_token outlines the default permissions.

.github/workflows/swift_package_test.yml
```diff
@@ -80,6 +84,10 @@ jobs:
         run: swift --version
       - name: Checkout repository
         uses: actions/checkout@v4
+      - name: Provide token
+        if: ${{ inputs.needs_token }}
+        run: |
+            echo "GITHUB_TOKEN=${{ secrets.GITHUB_TOKEN }}" >> $GITHUB_ENV
```
MahdiBM, line 90: I wonder if something like this at call sites would do what you want:
```diff
    uses: path/to/my/ci-file
+   secrets: inherit
```

award999: That didn't work unfortunately. secrets.GITHUB_TOKEN was empty

award999: Actually it wasn't the fact it was empty, secrets.GITHUB_TOKEN causes an error
Screenshot 2025-02-05 at 9 45 40 AM

and then github.secrets.GITHUB_TOKEN doesn't cause an error but is empty. And yes secrets is set to inherit

shahmishal (approved):

PR #108, opened 2025-03-15
Add support for macOS platform (default: false)

shahmishal:


.github/workflows/pull_request.yml
```diff
@@ -37,6 +37,20 @@ jobs:
         swift build
       enable_windows_docker: false
 
+  tests_macos:
```
ahoppen, line 40: Should this be part of `test_with_docker`? I would imagine that repos usually define one job that tests all the platforms. I would maybe rename `test_with_docker` to `Test` in that case and then we just have `test_without_docker` left to test Windows without Docker (and could probably rename to `test_windows_without_docker`.

.github/workflows/swift_package_test.yml
```diff
@@ -1,8 +1,24 @@
-name: Swift Linux Matrix
+name: Swift Matrix
 
 on:
   workflow_call:
     inputs:
+      macos_xcode_versions:
+        type: string
+        description: "Exclude Xcode version list (JSON)"
```
ahoppen, line 8:
```suggestion
        description: "Xcode version list (JSON)"
```

.github/workflows/swift_package_test.yml
```diff
@@ -1,8 +1,24 @@
...
+        default: "[\"16.0\", \"16.1\", \"16.2\"]"
+      macos_exclude_xcode_versions:
+        type: string
+        description: "Exclude Xcode version list (JSON)"
+        default: "[{\"xcode_version\": \"\"}]"
+      macos_versions:
+        type: string
+        description: "macOS version list (JSON)"
+        default: "[\"sequoia\"]"
```
ahoppen, line 17: I personally would have probably chosen the release version number (eg. `15` for Sequioa) instead of the marketing name but don’t have strong opinions here. Also just checking: We don’t want to add minor-version fidelity here (eg. the ability to select 15.4 instead of 15.0 or something like that)?

Finally, I think this assumes that we also have a self-hosted runner for this arch + macOS version combination, right? Maybe worth pointing that out in a comment, eg. something like `This requires a self-hosted macOS runner with this version to be set up.` 

If we don’t have near-term plans to support multiple OSs and arches, I would probably remove the option to not give users a false sense that they can test all macOS + arch + Xcode versions.

.github/workflows/swift_package_test.yml
```diff
@@ -1,8 +1,24 @@
...
+      macos_archs:
+        type: string
+        description: "macOS arch list (JSON)"
+        default: "[\"ARM64\"]"
```
ahoppen, line 21: Same comment as with the macOS version, I would add a comment like: `This requires a self-hosted macOS runner with this architecture to be set up.`

.github/workflows/swift_package_test.yml
```diff
@@ -60,6 +87,10 @@ on:
         type: boolean
         description: "Boolean to enable linux testing. Defaults to true"
         default: true
+      enable_macos_checks:
+        type: boolean
+        description: "Boolean to enable linux testing. Defaults to false"
```
ahoppen, line 92:
```suggestion
        description: "Boolean to enable macOS testing. Defaults to false"
```

ahoppen (approved):
Wohooooooo. Some comments below but nothing that would block this.

PR #109, opened 2025-03-21
Append swift-docc-plugin only when needed

fboemer:
Currently, the `docs-check:` workflow always adds the `swift-docc-plugin` dependency, forcing adopters to remove the dependency  from their package. This PR edits the docs check to skip adding the dependency when already present.

* A package with swift-docc-plugin: https://github.com/fboemer/swift-homomorphic-encryption/actions/runs/13997582365/job/39196407070?pr=3#step:5:222

* A package without swift-docc-plugin: https://github.com/fboemer/swift-homomorphic-encryption/actions/runs/13997685223/job/39196742443#step:5:222

### DIFF
--- a/.github/workflows/scripts/check-docs.sh
+++ b/.github/workflows/scripts/check-docs.sh
@@ -31,15 +31,23 @@ if [ -z "$package_files" ]; then
   fatal "Package.swift not found. Please ensure you are running this script from the root of a Swift package."
 fi
 
-for package_file in $package_files; do
-  log "Editing $package_file..."
-  cat <<EOF >> "$package_file"
+# yq 3.1.0-3 doesn't have filter, otherwise we could replace the grep call with "filter(.identity == "swift-docc-plugin") | keys | .[]"
+hasDoccPlugin=$(swift package dump-package | yq -r '.dependencies[].sourceControl' | grep -e "\"identity\": \"swift-docc-plugin\"" || true)
+if [[ -n $hasDoccPlugin ]]
+then
+    log "swift-docc-plugin already exists"
+else
+    log "Appending swift-docc-plugin"
+    for package_file in $package_files; do
+      log "Editing $package_file..."
+      cat <<EOF >> "$package_file"
 
 package.dependencies.append(
     .package(url: "https://github.com/swiftlang/swift-docc-plugin", "1.0.0"..<"1.4.0")
 )
 EOF
-done
+    done
+fi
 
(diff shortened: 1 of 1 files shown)


.github/workflows/scripts/check-docs.sh
```diff
@@ -31,15 +31,23 @@ if [ -z "$package_files" ]; then
   fatal "Package.swift not found. Please ensure you are running this script from the root of a Swift package."
 fi
 
-for package_file in $package_files; do
-  log "Editing $package_file..."
-  cat <<EOF >> "$package_file"
+# yq 3.1.0-3 doesn't have filter, otherwise we could replace the grep call with "filter(.identity == "swift-docc-plugin") | keys | .[]"
```
fboemer, line 34: * Trying out filter via `hasDoccPlugin=$(swift package dump-package | yq -r '.dependencies[].sourceControl | filter(.identity == "swift-docc-plugin") | keys | .[]')`
https://github.com/fboemer/swift-homomorphic-encryption/actions/runs/13997912746/job/39197486713
```
jq: error: filter/1 is not defined at <top-level>, line 1:[222](https://github.com/fboemer/swift-homomorphic-encryption/actions/runs/13997912746/job/39197486713#step:5:223)
.dependencies[].sourceControl | filter(.identity == "swift-docc-plugin") | keys | .[] 
jq: 1 compile error
yq: Error running jq: BrokenPipeError: [Errno 32] Broken pipe.
```

.github/workflows/scripts/check-docs.sh
```diff
@@ -31,15 +31,23 @@ if [ -z "$package_files" ]; then
...
+hasDoccPlugin=$(swift package dump-package | yq -r '.dependencies[].sourceControl' | grep -e "\"identity\": \"swift-docc-plugin\"" || true )
```
ahoppen, line 35: `swift package dump-package` is JSON, so should this be `jq` instead of `yq`?

Also: Trailing space after `true`)

ahoppen (commented):
Thanks for the contribution. Looks good to me, just one small comment.

.github/workflows/scripts/check-docs.sh
fboemer: We already use `yq` later on in the script, so I figured no need to an extra dependency, since `yq` supports json (as well as yaml)

ahoppen (approved):

FranzBusch (approved):

PR #110, opened 2025-04-01
Update Swift Package CI to include Swift 6.1 on Linux and Windows

shahmishal:


.github/workflows/swift_package_test.yml
```diff
@@ -73,7 +73,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        swift_version: ['5.9', '5.10', '6.0', 'nightly-main', 'nightly-6.1']
+        swift_version: ['5.9', '5.10', '6.0', '6.1', 'nightly-main', 'nightly-6.1']
```
FranzBusch, line 76: Should we drop 5.9 now? We normally only support the last 3 Swift versions across the server ecosystem. Maybe that's worth aligning over the entire ecosystem here.

shahmishal: Yes, but this will be something I will do in a different PR as we will need make an announcement on forums.swift.org. For example, https://forums.swift.org/t/dropping-support-for-swift-5-8-on-linux/78533

shahmishal:
Windows failures are expected as it's looking for scripts from `main` branch compared to pull request.

PR #111, opened 2025-04-04
Apply concurrency control to swift-package-test, cancelling previously running workflows of the same kind

TTOzzi:
Hello! 👋
This change is derived from https://github.com/swiftlang/swift-format/pull/976.
It adds concurrency control to ensure that only one active workflow is running per pull request, preventing duplicate workflow executions for open PRs.

### DIFF
--- a/.github/workflows/swift_package_test.yml
+++ b/.github/workflows/swift_package_test.yml
@@ -65,6 +65,11 @@ on:
         description: "Boolean to enable providing the GITHUB_TOKEN to downstream job."
         default: false
 
+## We are cancelling previously triggered workflow runs
+concurrency:
+  group: ${{ github.workflow }}-${{ github.job }}-${{ github.ref }}-package-test
+  cancel-in-progress: true
+
 jobs:
   linux-build:
     name: Linux (${{ matrix.swift_version }} - ${{ matrix.os_version }})


TTOzzi:
It seems that in workflows like [this one](https://github.com/swiftlang/github-workflows/blob/main/.github/workflows/pull_request.yml), where multiple `swift_package_test` jobs are executed, they share the same concurrency group ID, which causes them to cancel each other 🤔

TTOzzi:
Currently, when the `swift_package_test` workflow is used by multiple jobs within a parent workflow, it cannot access information about the specific job that called it. As a result, the concurrency group ID becomes the same across jobs, causing them to be treated as duplicates and canceled.

I’ve tried several approaches based on [GitHub’s documentation](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs), but so far, it seems the only way to avoid this is to explicitly pass a unique identifier from each job in the parent workflow to the `swift_package_test`.

However, requiring such an ID as an input parameter would affect all existing workflows that use `swift_package_test`, essentially defeating the purpose by shifting the concurrency logic back to each individual job.

TTOzzi:
As long as there are cases where swift_package_test is called from multiple jobs within the same workflow, it seems difficult to handle concurrency in a centralized way 😞 
I'll go back to the original PR.

.github/workflows/swift_package_test.yml
```diff
@@ -65,6 +65,11 @@ on:
         description: "Boolean to enable providing the GITHUB_TOKEN to downstream job."
         default: false
 
+## We are cancelling previously triggered workflow runs
+concurrency:
+  group: ${{ github.workflow }}-${{ github.job }}-${{ github.ref }}-package-test
```
TTOzzi, line 70: While I could tell which PR it was and whether it was the `linux-build` or `windows-build` job inside `swift_package_test`, there was no way — using only the information available within `swift_package_test` — to determine whether it was triggered from `tests_with_docker` or `tests_without_docker` in the [parent workflow](https://github.com/swiftlang/github-workflows/blob/main/.github/workflows/pull_request.yml).

PR #128, opened 2025-05-15
Add support for `nightly-6.2` on Linux in `swift_package_test.yml`

MaxDesiatov:
Since 6.2 development snapshots are already available, it's worth including those in the Linux build matrix


MaxDesiatov:
Windows failure seems unrelated, but I can't re-trigger it to check if it's deterministic.

FranzBusch (approved):

PR #129, opened 2025-05-29
Retry curl operations in CI scripts

rnro:
Some 429: Too many requests responses have been seen when attempting to curl remote CI scripts.

This change adds `--retry 3` which means that curl will retry 3 times on transient error codes, first for 1 second and then doubled each retry.

from `man curl`
```
       --retry <num>
              If a transient error is returned when curl tries to perform a transfer, it
              retries this number of times before giving up. Setting the number to 0
              makes curl do no retries (which is the default). Transient error means
              either: a timeout, an FTP 4xx response code or an HTTP 408, 429, 500, 502,
              503 or 504 response code.

              When curl is about to retry a transfer, it first waits one second and then
              for all forthcoming retries it doubles the waiting time until it reaches 10
              minutes which then remains delay between the rest of the retries. By using
              --retry-delay you disable this exponential backoff algorithm. See also
              --retry-max-time to limit the total time allowed for retries.

              curl complies with the Retry-After: response header if one was present to
              know when to issue the next retry (added in 7.66.0).

              If --retry is provided several times, the last set value is used.
```

### DIFF
--- a/.github/workflows/soundness.yml
+++ b/.github/workflows/soundness.yml
@@ -144,7 +144,7 @@ jobs:
           ADDITIONAL_DOCC_ARGUMENTS: ${{ inputs.docs_check_additional_arguments }}
         run: |
           which curl yq || (apt -q update && apt -yq install curl yq)
-          curl -s https://raw.githubusercontent.com/swiftlang/github-workflows/refs/heads/main/.github/workflows/scripts/check-docs.sh | bash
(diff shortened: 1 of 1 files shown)


shahmishal:
Not sure if it will help but we can try. Normally `429` return code is for rate limit.

shahmishal (approved):

PR #134, opened 2025-06-16
Add release-mode test runs for macOS and Linux

natecook1000:
This adds a `configuration: [debug, release]` element to the test matrices for macOS and Linux so that tests can be run in both modes.


rauhul:
Why not just run in release always?

FranzBusch:
> Why not just run in release always?

Most packages don't support building tests in release mode due to `@testable` imports AFAIK

rauhul:
`@testable` supports release mode. It just makes `internal` api visible outside the module unit.

jrflat:
FWIW even with `--enable-testable-imports`, System was still failing to build tests in release mode (due to some `#if DEBUG` conditionals, fixed by PR above). We would probably need to keep `enable_release_tests` defaulting to `false` or otherwise track down/help fix potential failures in projects using this workflow.

MahdiBM:
The way to enable @testable imports even in release builds is to use ([`-Xswiftc -enable-testing`](https://github.com/vapor/ci/blob/fc0c1f160e488f0affd1d03508db5bd76c1f7ce3/.github/workflows/run-unit-tests.yml#L99)) like I've added a few months ago to Vapor's reusable CI for release-mode testing of packages when possible.
I'm not quite sure why `--enable-testable-imports` doesn't cut it. Could be a bug.

PR #141, opened 2025-07-30
Fix the Swift versions list for Windows

glessard:
5.10 was omitted. This adds a comment explaining the omission.

### DIFF
--- a/.github/workflows/swift_package_test.yml
+++ b/.github/workflows/swift_package_test.yml
@@ -34,7 +34,8 @@ on:
       windows_swift_versions:
         type: string
         description: "Include Windows Swift version list (JSON)"
-        default: "[\"5.9\", \"5.10\", \"6.0\", \"6.1\", \"nightly\", \"nightly-6.2\"]"
+        # "5.10" is omitted for Windows because the container image is broken.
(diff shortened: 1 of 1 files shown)


award999:
Ya I'd say that the build output looks off https://github.com/swiftlang/github-workflows/actions/runs/16629101032/job/47053544826?pr=141#step:9:65. Not sure why that didn't fail the verify job though

award999:
Output from the other Windows docker images look fine in the verifier jobs but I would not merge this

glessard:
Is there a way to massage the absence so that it doesn't look like an omission?
It simply doesn't look intentional.

ahoppen:
Could you open a PR to add a comment explaining what we said above? Or maybe re-purpose this one?

ahoppen (approved):
Thank you. LGTM.

bnbarham:
> Ya I'd say that the build output looks off https://github.com/swiftlang/github-workflows/actions/runs/16629101032/job/47053544826?pr=141#step:9:65

I don't know why it *was* failing. But it would definitely be failing now because of the missing SDK in the latest build tools. Not sure if that was also the same reason, `invalid manifest` matches that, though I would have expected a little more output (ie. for the missing libraries, as in https://github.com/apple/swift-numerics/actions/runs/16943623923/job/48018951798). Do we have any issue to add 5.10 back in?

PR #145, opened 2025-08-04
Add Embedded Swift for Wasm support to `install-and-build-with-sdk.sh`

MaxDesiatov:
Also cleaned up incorrect "SDK" naming: we're using Swift SDKs and not plain Clang or Xcode SDKs in these workflows.

Additionally "static SDK" is very non-specific, Swift SDKs for Wasm are also static, as we currently don't support dynamic linking on that platform. We should fully specify it as "Static Linux", same as in https://www.swift.org/documentation/articles/static-linux-getting-started.html

### DIFF
--- a/.github/workflows/scripts/install-and-build-with-sdk.sh
+++ b/.github/workflows/scripts/install-and-build-with-sdk.sh
@@ -203,7 +203,7 @@ if [[ "$SWIFT_VERSION_INPUT" == nightly-* ]]; then
     if [[ "$INSTALL_STATIC_LINUX" == true ]]; then
         static_linux_sdk_info=$(find_latest_sdk_snapshot "$version" "static")
 
-        STATIC_LINUX_SDK_TAG=$(echo "$static_linu_sdk_info" | cut -d'|' -f1)
+        STATIC_LINUX_SDK_TAG=$(echo "$static_linux_sdk_info" | cut -d'|' -f1)
         STATIC_LINUX_SDK_CHECKSUM=$(echo "$static_linux_sdk_info" | cut -d'|' -f2)
     fi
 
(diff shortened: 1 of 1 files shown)


.github/workflows/scripts/install-and-build-with-sdk.sh
```diff
@@ -112,11 +118,11 @@ find_latest_swift_version() {
...
         ')
 
-        if [[ -z "$static_checksum" ]]; then
-            fatal "No static SDK checksum found for Swift $latest_version"
+        if [[ -z "$static_linux_checksum" ]]; then
+            fatal "No Static Linux Swift SDK checksum found for Swift $latest_version"
         fi
 
-        log "Found static SDK checksum: ${STATIC_SDK_CHECKSUM:0:12}..."
+        log "Found Static Linux Swift SDK checksum: ${STATIC_LINUX_SDK_CHECKSUM:0:12}..."
```
jrflat, line 125: Oops I made a mistake in the original script when refactoring, could you change this to the lowercase `static_linux_checksum`?

MaxDesiatov:
> Thanks for fixing the SDK terminology and adding embedded support! Left some comments on logs that could also be updated from just "static SDK" but looks good.

Thanks!

> Just to clarify, the `_wasm-embedded` SDK is included in the `_wasm.artifactbundle.tar.gz` that's installed, right?

Yes. 
 
> Also, do we want to add a job that tests the embedded build in `swift_package_test.yml` and `pull_request.yml`?

Those will be changed in follow up PRs, as I don't see a way for a workflow change PR to refer to a script in the same, as before merging we need to refer to `main` and PR branch is not `main`.

.github/workflows/scripts/install-and-build-with-sdk.sh
jrflat: Just there since it should be the local `static_linux_checksum` we just found and `STATIC_LINUX_SDK_CHECKSUM` isn't defined yet.

MaxDesiatov: Makes sense, that's addressed now!

Issue #146, opened 2025-08-04
Linux Static SDK and WebAssembly not working with RPM distros


jakepetroules:
The new CI jobs introduced by #142 are hardcoded to use `apt`. Please add a mechanism that tries both `apt` and `dnf` to install dependent packages so that these workflows work on RPM distros like RHEL UBI and Fedora.

jakepetroules:
@jrflat could you take a look?

jrflat:
That's a good idea will do!

PR #148, opened 2025-08-04
Support RPM distros when building with Swift static Linux/Wasm SDKs

jrflat:
Detect package manager (`apt`, `dnf`, or `yum`) to use for installing dependencies. Also, update `initialize_os_info()` to use the expected OS name for the download URL (such as `ubi9`), which might not match the `ID` and `VERSION_ID` in `/etc/os-release`.

Verified the script now works on rhel-ubi9, fedora39, amazonlinux2, and debian12 containers. This testing also revealed an issue where the script would continue running and fail if the required toolchain couldn't be found, rather than exiting cleanly as intended. I fixed this so the script exits successfully and doesn't fail the CI job in this case.

Resolves https://github.com/swiftlang/github-workflows/issues/146

### DIFF
--- a/.github/workflows/scripts/install-and-build-with-sdk.sh
+++ b/.github/workflows/scripts/install-and-build-with-sdk.sh
@@ -73,9 +73,24 @@ if [[ -n "$SWIFT_BUILD_FLAGS" ]]; then
     log "Additional build flags: $SWIFT_BUILD_FLAGS"
 fi
 
+# Detect package manager
+if command -v apt >/dev/null 2>&1; then
+    INSTALL_PACKAGE_COMMAND="apt update -q && apt install -yq"
+elif command -v dnf >/dev/null 2>&1; then
+    INSTALL_PACKAGE_COMMAND="dnf install -y"
+elif command -v yum >/dev/null 2>&1; then
+    INSTALL_PACKAGE_COMMAND="yum install -y"
+else
+    fatal "No supported package manager found"
+fi
+
+install_package() {
+    eval "$INSTALL_PACKAGE_COMMAND $1"
+}
+
 # Install dependencies
-command -v curl >/dev/null || (apt update -q && apt install -yq curl)
-command -v jq >/dev/null || (apt update -q && apt install -yq jq)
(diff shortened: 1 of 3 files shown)


shahmishal (approved):

.github/workflows/scripts/install-and-build-with-sdk.sh
```diff
@@ -73,9 +73,24 @@ if [[ -n "$SWIFT_BUILD_FLAGS" ]]; then
     log "Additional build flags: $SWIFT_BUILD_FLAGS"
 fi
 
+# Detect package manager
+if command -v apt >/dev/null 2>&1; then
+    INSTALL_PACKAGE_COMMAND="apt update -q && apt install -yq"
+elif command -v dnf >/dev/null 2>&1; then
+    INSTALL_PACKAGE_COMMAND="dnf install -y"
+elif command -v yum >/dev/null 2>&1; then
+    INSTALL_PACKAGE_COMMAND="yum install -y"
```
jakepetroules, line 82: nit: the yum branch isn't really needed. yum has been replaced by dnf, and the only distro supported by Swift which hasn't fully moved yet is Amazon Linux 2, but that distro doesn't even work in GitHub actions in general because a number of its constituent components are too old for GitHub to set up its communication pipe

jakepetroules (commented):
Thanks for landing this so quickly!

.github/workflows/scripts/install-and-build-with-sdk.sh
jrflat: Ah that makes sense, I was using `yum` when testing locally on AL2 so I included it, but I didn't realize it wasn't supported for GitHub actions. I think I'll probably keep it in the script for now in case there's a need to use it locally in an AL2 container, e.g. running

```
curl -s --retry 3 https://raw.githubusercontent.com/swiftlang/github-workflows/refs/heads/main/.github/workflows/scripts/install-and-build-with-sdk.sh | \
bash -s -- [--static] [--wasm] [--embedded-wasm] --flags="<build-flags>" <swift-version>
```

PR #149, opened 2025-08-05
Support custom build commands for Swift static Linux/Wasm SDK builds

### LINKED ISSUE #147: Linux Static SDK and WebAssembly build commands are not overridable
The new CI jobs introduced by #142 are hardcoded to use a fixed build command. Please add a property that can be used to override this -- for the Static Linux SDK in particular there is no way to distinguish it from regular Linux builds in the package manifest, so it's often useful to be able to pass an environment variable to control the build in cases where it must differ.

jrflat:
Enable alias expansion and run `alias swift="$SWIFT_EXECUTABLE_FOR_(NAME)_SDK"` before performing the build so that we can support custom `swift build` commands while still using the correct executable.

I'll add a commit that points to the script URL on my branch for testing this PR, then remove it before merging.

Resolves #147

### DIFF
--- a/.github/workflows/swift_package_test.yml
+++ b/.github/workflows/swift_package_test.yml
@@ -242,7 +242,7 @@ jobs:
         run: |
           ${{ inputs.linux_static_sdk_pre_build_command }}
           which curl || (apt -q update && apt -yq install curl)
-          curl -s --retry 3 https://raw.githubusercontent.com/jrflat/github-workflows/refs/heads/custom-command-for-sdk-builds/.github/workflows/scripts/install-and-build-with-sdk.sh | \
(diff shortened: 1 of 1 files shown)


shahmishal (approved):
Looks good, make sure to review the forked repo path once test is completed. 

- [ ] Revert the script back to swiftlang/github-workflows url

jrflat:
Reverted the script URL back to `swiftlang:main`. Previous successful run for reference: https://github.com/swiftlang/github-workflows/actions/runs/16738403255

jrflat:
This is going to conflict with the name/description changes in #150. I'll wait for that to land or remove my changes that'll conflict.

shahmishal (approved):

PR #151, opened 2025-08-14
Fix support for amazonlinux2 with the swift_package_test workflow

jakepetroules:
checkout@v4 requires a version of node that requires a newer version of glibc than exists on this OS version, so downgrade it to v3 for AL2.


justice-adams-apple:
lgtm pending green ci

PR #152, opened 2025-08-14
Use v3 checkout with amazonlinux2

justice-adams-apple:


jakepetroules:
This should be closed as it was superseded by #151.

PR #153, opened 2025-08-15
Bump the default minimum Swift version to 5.9 in package tests

bnbarham:


bnbarham:
Seems the build tools fix did not fix 5.10 :(

PR #154, opened 2025-08-20
Do not assume windows_build_command is a single line

bnbarham:
`windows_build_command` was prefixed with a `Invoke-Program` to ensure we catch errors, but this then misses multiline commands. Move `Invoke-Program` up into the default command and then assume that it is passed if a non-default command is used.

Note that `$ErrorActionPreference = 'Stop'` is not an option here - it only stops for cmdlets, not native commands.
`$PSNativeCommandUseErrorActionPreference = $true` allows stopping on native commands too, but that was only introduced in PowerShell 7.3. windowsservercore-ltsc2022 has 5.1 by default (though we could install a newer if we wanted to).

Update PR testing in this repo to use `Invoke-Program` to then actually catch failures.

### DIFF
--- a/.github/workflows/swift_package_test.yml
+++ b/.github/workflows/swift_package_test.yml
@@ -102,7 +102,7 @@ on:
           Windows Command Prompt command to build and test the package.
           Note that Powershell does not automatically exit if a subcommand fails. The Invoke-Program utility is available to propagate non-zero exit codes.
           It is strongly encouraged to run all command using `Invoke-Program` unless you want to continue on error eg. `Invoke-Program git apply patch.diff` instead of `git apply patch.diff`.
-        default: "swift test"
+        default: "Invoke-Program swift test"
       macos_env_vars:
         description: "Newline separated list of environment variables"
         type: string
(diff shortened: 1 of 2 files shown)


.github/workflows/swift_package_test.yml
```diff
@@ -450,7 +450,7 @@ jobs:
           Invoke-Program swift test --version
           Invoke-Program cd $Source
           ${{ inputs.windows_pre_build_command }}
-          Invoke-Program ${{ inputs.windows_build_command }} ${{ (contains(matrix.swift_version, 'nightly') && inputs.swift_nightly_flags) || inputs.swift_flags }}
```
bnbarham, line 453: We could also leave this in, I am slightly concerned that `Invoke-Program` isn't being used in downstream overrides today.

ahoppen: I don’t think we should prefix it with `Invoke-Program`. Otherwise, you might learn that the command fails if you only have a single one in your `windows_build_commands` and are then surprised that this behavior doesn’t hold up when you add a second command to the script. 

Instead, what I would suggest is that we go through the existing use cases of this workflow with `windows_build_command` and fix them. There’s only a handful and many of them are forks: https://github.com/search?q=swiftlang%2Fgithub-workflows%2F.github%2Fworkflows%2Fswift_package_test.yml%40main+windows_build_command&type=code

ahoppen (approved):
LGTM but we should fix all existing innovations of this workflow to use `Invoke-Program`.

.github/workflows/swift_package_test.yml
bnbarham: Yeah, this is why I removed it. I'll go through them

bnbarham: https://github.com/swiftlang/swift-syntax/pull/3140
https://github.com/swiftlang/swift-build/pull/746
https://github.com/swiftlang/vscode-swift/pull/1800
https://github.com/swiftlang/swift-foundation-icu/pull/77
https://github.com/swiftlang/swift-format/pull/1060
https://github.com/swiftlang/swift-toolchain-sqlite/pull/20

PR #159, opened 2025-09-16
Add Swift 6.2 version to swift package testing

shahmishal:


award999:
@shahmishal I added #160. If you want to merge that PR and rebase this patch, then the 6.2 without docker job should then pass

PR #164, opened 2025-09-22
Update container images for soundness to 6.2

FranzBusch:


FranzBusch:
macOS failure is expected and unrelated

PR #176, opened 2025-10-28
De-matrix Android triples

jakepetroules:
Building for multiple ABIs at once is something the swift-build frontend does for macOS and should eventually do for other platforms. Don't "lift" the triple into a new matrix axis.

This also renames the android_sdk_triple and android_ndk_version properties to plurals, since they support multiple versions. This is breaking, but the Android workflow is only a couple of days old, so this should be fine.

### DIFF
--- a/.github/workflows/swift_package_test.yml
+++ b/.github/workflows/swift_package_test.yml
@@ -528,6 +528,26 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v1
         if: ${{ matrix.os_version == 'amazonlinux2' }}
+      - name: Checkout swiftlang/github-workflows repository
+        if: ${{ matrix.os_version != 'amazonlinux2' && github.repository != 'swiftlang/github-workflows' }}
+        uses: actions/checkout@v4
+        with:
+          repository: swiftlang/github-workflows
+          path: github-workflows
+      - name: Checkout swiftlang/github-workflows repository
+        if: ${{ matrix.os_version == 'amazonlinux2' && github.repository != 'swiftlang/github-workflows' }}
+        uses: actions/checkout@v1
+        with:
+          repository: swiftlang/github-workflows
+          path: github-workflows
+      - name: Determine script-root path
+        id: script_path
+        run: |
+          if [ "${{ github.repository }}" = "swiftlang/github-workflows" ]; then
+            echo "root=$GITHUB_WORKSPACE" >> $GITHUB_OUTPUT
+          else
+            echo "root=$GITHUB_WORKSPACE/github-workflows" >> $GITHUB_OUTPUT
+          fi
       - name: Provide token
         if: ${{ inputs.needs_token }}
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/swift_package_test.yml
```diff
@@ -511,8 +511,7 @@ jobs:
       fail-fast: false
       matrix:
         swift_version: ${{ fromJson(inputs.android_sdk_versions) }}
-        sdk_triple: ${{ fromJson(inputs.android_sdk_triple) }}
```
justice-adams-apple: Question about removing the triple from the matrix axis: does this make it harder for end users to diagnose build issues in a particular architecture?

Just thinking from the concept of a CI job, if a dev were testing their package and wanted to test against the Swift SDK for Android, I would think it would create a job run per triple, but admittedly I haven't though too deeply about it: So I guess the outstanding question is just what the benefit is from a CI users perspective

.github/workflows/swift_package_test.yml
```diff
@@ -529,6 +528,26 @@ jobs:
...
         if: ${{ matrix.os_version == 'amazonlinux2' }}
+      - name: Checkout swiftlang/github-workflows repository
+        if: ${{ matrix.os_version != 'amazonlinux2' && github.repository != 'swiftlang/github-workflows' }}
+        uses: actions/checkout@v4
+        with:
+          repository: swiftlang/github-workflows
+          path: github-workflows
+      - name: Checkout swiftlang/github-workflows repository
+        if: ${{ matrix.os_version == 'amazonlinux2' && github.repository != 'swiftlang/github-workflows' }}
+        uses: actions/checkout@v1
```
justice-adams-apple: could we pin this to a specific revision, like `main`?
```ref: 'main' ```

I don't want to run the risk of running the workflow in this repository (say as part of a CI test) and having this check out a users branch which could contain a malicious version of the script being executed

.github/workflows/swift_package_test.yml
justice-adams-apple: makes sense, thanks for explaining 👍

.github/workflows/swift_package_test.yml
justice-adams-apple: After re-reading the security policy about approving workflow runs
https://docs.github.com/en/actions/how-tos/manage-workflow-runs/approve-runs-from-forks

This might not be a real concern, since we have to approve all PR workflow runs from non-write-access-contributors.

 I'll defer to @shahmishal on this one

jakepetroules: 'ref' defaults to the default branch of the repository, so I think this would not do anything interesting. Furthermore, someone could just put the malicious script inline in the YAML instead of referencing an external one by URL. You're right that workflow approvals are the real security tool here.

.github/workflows/swift_package_test.yml
jakepetroules: This changes the default from x86_64 only to x86_64 + aarch64, and the overhead will of course be proportional to the package size. Mainly I wanted to ensure the multiarch support worked correctly in testing this PR, but I think it's beneficial to have both on by default since aarch64 is what's used for the vast majority of devices, and x86_64 for the majority of at-desk testing in the emulator.

Could also revert to a single arch if you strongly prefer.

marcprux:
FWIW, LGTM

PR #177, opened 2025-11-03
Retry installing Visual Studio Build Tools on failure

plemarquand:
We've seen occasional failures to install the Visual Studio Build Tools due to flakey responses from the server.

Attempt to make this script more robust by adding a retry with exponential backoff, attempting up to 10 times before eventually giving up. This adds the same logic to both the VSB and Swift toolchain downloads.


.github/workflows/scripts/windows/web-request-utils.psm1
```diff
@@ -0,0 +1,75 @@
+##===----------------------------------------------------------------------===##
+##
+## This source file is part of the Swift.org open source project
+##
+## Copyright (c) 2024 Apple Inc. and the Swift project authors
```
award999: 2025

jakepetroules:
Is there any advantage to using Invoke-WebRequest versus just using curl with --retry? Windows has it built-in nowadays. The code would be a lot simpler.

jakepetroules (approved):

PR #186, opened 2025-11-11
Update github-workflows source location for Windows docker

bnbarham:
The workspace is mounted into `C:\source` when using docker, update the script root to take that into account.

### DIFF
--- a/.github/workflows/swift_package_test.yml
+++ b/.github/workflows/swift_package_test.yml
@@ -684,10 +684,16 @@ jobs:
       - name: Determine script-root path
         id: script_path
         run: |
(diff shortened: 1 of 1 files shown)


bnbarham:
Fixes https://github.com/swiftlang/swift-format/actions/runs/19255344965/job/55048760484?pr=1086, which has been broken since https://github.com/swiftlang/github-workflows/pull/181. CC @jakepetroules 

We could also create yet another output since we do the `if ("${{ inputs.enable_windows_docker }}" -eq "true") {` check a little further down as well 🤷‍♂️

shahmishal (approved):

PR #188, opened 2025-11-19
Various fixes for Amazon Linux 2

jakepetroules:
- Check out github-workflows into subdirectory: checkout@v1 and checkout@v4 differ in their behavior, try to make v1 clone the workflows repo in the same place as v4 so that on Amazon Linux 2 it doesn't try to run the script from the wrong location
- Don't use unnecessary piping: this causes script execution failures to be ignored when -o pipefail is not set.
- Add missing executable bit to install-and-build-with-sdk.sh script
- Don't install curl unnecessarily in order to save time (this became unnecessary with #175, #180, #181, and #182)


jakepetroules:
I updated the description with the commit comments.

PR #195, opened 2025-12-02
Add option for automerger workflow to create non-draft PRs

bkhouri:
Add an option to the GitHub automerger workflow to optionally create a PR as non-draft.  When this is the case, the tests will be demanded by commenting `@swift-ci please test` on the created PR.

### DIFF
--- a/.github/workflows/create_automerge_pr.yml
+++ b/.github/workflows/create_automerge_pr.yml
@@ -30,16 +30,24 @@ name: Create automerge PR
 #       pull-requests: write
 #     with:
 #       base_branch: release/6.2
+#       # Optional: Set to false to automatically trigger tests via "@swift-ci please test" comment
+#       # create_as_draft: false
 # ```
 #
 # PRs created by GitHub Actions don't kick off further actions (https://github.com/peter-evans/create-pull-request/blob/d57e551ebc1a16dee0b8c9ea6d24dba7627a6e35/docs/concepts-guidelines.md#triggering-further-workflow-runs).
-# As a workaround, we mark automerge PRs that are created by GitHub actions as draft and trigger the GitHub actions by marking the PR as ready for review. `ready_for_review` must be added to the PR types for this purpose, eg.
-# ```
-# on:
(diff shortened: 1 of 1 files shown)


.github/workflows/create_automerge_pr.yml
```diff
@@ -102,7 +118,13 @@ jobs:
...
             --title 'Merge `${{ inputs.head_branch }}` into `${{ inputs.base_branch }}`' \
-            --body '${{ inputs.pr_message }}' \
-            --draft
+            --body '${{ inputs.pr_message }}'
+
+          # Trigger tests is requested
+          if [[ "${{ inputs.create_as_draft }}" != "true" ]] ; then
+            PR_URL=$(gh pr view "${PR_BRANCH}" --json url | jq -r '.url')
+
+            gh pr comment "$PR_URL" --body "@swift-ci please test"
```
jakepetroules: nit: just `@swift-ci test` works

.github/workflows/create_automerge_pr.yml
```diff
@@ -30,16 +30,24 @@ name: Create automerge PR
...
+#    For this approach, `ready_for_review` must be added to the PR types, eg.
+#    ```
+#    on:
+#     pull_request:
+#       types: [..., ready_for_review]
+#    ```
+#    Unfortunately this will also re-trigger testing even on a normal user's PR (which may have already been tested), but skipping them causes the checks to reset so this is the best we can do for now.
+#
+# 2. **Auto-trigger approach**: When `create_as_draft` is set to `false`, the workflow automatically triggers tests by commenting "@swift-ci please test" on the created PR.
+#    This requires the repository to have the swift-ci bot configured to respond to such comments.
```
ahoppen: We should point out that this does not trigger any of the GitHub action testing, just Swift CI testing.

.github/workflows/create_automerge_pr.yml
bkhouri: I know, but let's encourage folks to be polite :)

PR #198, opened 2025-12-02
Add retry option to curl commands in install script

shahmishal:


.github/workflows/scripts/install-and-build-with-sdk.sh
```diff
@@ -400,16 +400,16 @@ download_and_verify() {
     local temp_sig="${output_file}.sig"
 
     log "Downloading ${url}"
```
shahmishal:
```suggestion
    log "Downloading with retry flag ${url}"
```

PR #172, opened 2025-10-21
Add Swift SDK for Android to swift_package_test workflow

marcprux:
Pursuant to https://github.com/swiftlang/github-workflows/issues/83, this PR adds support for building with the Swift SDK for Android, following the precedents set by the Swift SDK for Wasm and the Static Linux SDK.

Unlike https://github.com/swiftlang/github-workflows/pull/106, this PR doesn't cover testing, which would necessitate installing and configuring an Android emulator.

Issue #83, opened 2025-01-06
Add an Android workflow, for Swift packages to build the Android SDK and test against


finagolfin:
I've had requests for an Android workflow, both using my unofficial SDK, finagolfin/swift-android-sdk#127, and for [whenever an official Android SDK is available](https://github.com/apple/swift-nio/pull/2784#pullrequestreview-2174034777). Rather than waiting for an official Android SDK first, I thought I could put together an Android workflow here that builds the Android SDK from source on first use, then uses it to cross-compile that Swift package that invoked this workflow on its CI and run its tests in the Android emulator, before finally caching the freshly-built Android SDK on that Swift package's CI for subsequent test runs.

Here are some further details:
- I'd start off with Swift 6.1 only, building from the latest snapshot tag, then add other release branches later.
- linux-only for build host initially, add macOS later
- I currently download [some pre-compiled libraries for Android from another OSS project](https://github.com/finagolfin/swift-android-sdk/blob/bb1cef608cc8c327705f70253479c1a0ccffe803/get-packages-and-swift-source.swift#L6) for my SDK bundle supporting Android API 24: a backported `posix_spawn()` wrapper to Android API 24 and libcurl/libxml2 for Foundation. I'd simply require Android API 28, when `posix_spawn()` was added to Android, and remove `FoundationNetworking` and `FoundationXML` support for this workflow, looking into ways to add those back later, so no such external libraries would be required for this workflow.
- I'd simply use an existing external Android emulator workflow, [as I do on my Android CI](https://github.com/finagolfin/swift-android-sdk/blob/main/.github/workflows/sdks.yml#L528). We could look into removing that external dependency later, if wanted.
- Once an official Android SDK is available, we could make that the default, and make this fresh build an alternate option for those who want it.

@FranzBusch or @shahmishal, let me know if you'd be okay with adding such an Android workflow here, will put together a pull if so.

marcprux:
We might want to base this on https://github.com/marketplace/actions/swift-android-action, which supports building and testing on both Linux and macOS (Intel) hosts, as well as caching the SDK, host toolchain, and emulator.

The SDK is currently downloaded from a mirror of https://github.com/finagolfin/swift-android-sdk/releases, but adding a step to optionally build it from source would be a possibility, at least for Linux.

finagolfin:
Fine by me, if you've done the work already and want to submit it here.

marcprux:
> if you've done the work already [...]

[...]

PR #215, opened 2025-12-15
Add enable_android_sdk_checks option for Android emulator testing

### LINKED ISSUE #179: Add Android emulator testing to workflow
With https://github.com/swiftlang/github-workflows/pull/172, we now have the option to build packages for Android from CI. But we don't yet have the ability to run tests.

Performing tests would involve downloading and configuring an Android emulator, ideally customizable with parameters like the Android API level to test. This sort of thing is handled nicely by the venerable [android-emulator-runner](https://github.com/reactivecircus/android-emulator-runner) action, and it is what my [swift-android-action](https://github.com/marketplace/actions/swift-android-action) uses for testing against Android, but my understanding is that the swiftlang workflows don't want to rely on external actions. So we'd need to manage the setup of the emulator myself. I've put together a sample workflow that does this at https://github.com/swift-android-sdk/swift-android-sdk-workflow-demo/blob/main/.github/workflows/ci.yml to show how this might be accomplished.

The other issue is the complexity of running tests on a separate "host" that doesn't share the same file system. The cross-compiled test binary needs to be transferred over to the QEMU emulator, along with any resources that it depends on. This logic gets hairy, and sometimes requires special parameters to identify special files that need copying over (e.g., SwiftCrypto looks for some test files [relative to the test's `#filePath` path](https://github.com/apple/swift-crypto/blob/e8ed8867ec23bccf5f3bb9342148fa8deaff9b49/Tests/CryptoExtrasTests/OPRFs/VOPRFAPITests.swift#L27) … this won't work without modifying the test case to look elsewhere).

Finally, to support testing both XCTest and Swift Testing tests, there's some fancy footwork required, as discussed at https://github.com/swiftlang/swift-package-manager/issues/8362#issuecomment-2719563933. You can see how I do it at `swift-android-action`'s [action.yml](https://github.com/skiptools/swift-android-action/blob/9ce927705c28d836e1d5c2a13326cf55e2c0851e/action.yml#L567).

All in all, it is achievable and I think it would be worthwhile for the benefit of actually exercising the tests on an Android emulator rather than just seeing if the source compiles.

marcprux:
The Android CI added in https://github.com/swiftlang/github-workflows/pull/172 will build against the Swift SDK for Android using the `scripts/install-and-build-with-sdk.sh` support for building with cross-compilation SDKs, but it does't actually test the package. This PR adds a `scripts/android/android-emulator-tests.sh` script that will install and configure an Android emulator and then transfer the test cases over to it and run the tests on the emulator.

Notes:

- In addition to the build-only `enable_android_sdk_build` input to `workflows/swift_package_test.yml`, we now have an `enable_android_sdk_checks` boolean which will additionally run a package's test cases.
- The Android emulator requires KVM to be enabled for acceleration, which is not possible in a nested docker container. I saw no reason why the SDK build needs to take place in a nested container, so I moved it to run on the top-level ubuntu-latest container.
- The emulator will only run on x86_64; since we support building against multiple triples (currently defaulting to `aarch64-unknown-linux-android28` and `x86_64-unknown-linux-android28` in `workflows/swift_package_test.yml`), we will still build against each of these archs, but only test against `x86_64`.
- As mentioned previously (https://github.com/swiftlang/github-workflows/issues/179#issuecomment-3492941426), not all package testing will pass out of the box, since some tests rely on hardcoded `#filePath` references to test data (e.g., in [swift-crypto](https://github.com/apple/swift-crypto/blob/6f70fa9eab24c1fd982af18c281c4525d05e3095/Tests/CryptoTests/Utils/RFCVector.swift#L32)), which will not work when the tests are run in an emulator with a separate filesystem. I expect that some of these packages will need to be updated if we are to enable `enable_android_sdk_checks` for them.

A successful run can be seen at https://github.com/swift-android-sdk/github-workflows/actions/runs/20244304679/job/58120240667

Closes https://github.com/swiftlang/github-workflows/issues/179 and https://github.com/swiftlang/github-workflows/issues/83

CC: @finagolfin, @justice-adams-apple, @etcwilde, @shahmishal


.github/workflows/scripts/android/android-emulator-tests.sh
```diff
@@ -0,0 +1,140 @@
...
+
+# extract the API level from the end of the triple 
+ANDROID_API="${ANDROID_SDK_TRIPLE/*-unknown-linux-android/}"
+
+# extract the build arch from the beginning of the triple
+ANDROID_EMULATOR_ARCH="${ANDROID_SDK_TRIPLE/-unknown-linux-android*/}"
+
+# x86_64=x86_64, armv7=arm
+ANDROID_EMULATOR_ARCH_TRIPLE="${ANDROID_EMULATOR_ARCH}"
+
```
finagolfin: Stick a log message here saying which Android arch is being tested.

marcprux: Added in 15a740d35e7f058cc73ab30a2b80ed251c9054fd

.github/workflows/scripts/android/android-emulator-tests.sh
```diff
@@ -0,0 +1,140 @@
...
+TEST_CMD="./*.xctest"
+TEST_SHELL="cd ${ANDROID_TMP_FOLDER}"
+TEST_SHELL="${TEST_SHELL} && ${TEST_CMD} --testing-library xctest"
+
+# Run test cases a second time with the Swift Testing library
+# We additionally need to handle the special exit code
+# EXIT_NO_TESTS_FOUND (69 on Android), which can happen
+# when the tests link to Testing, but no tests are executed
+# see: https://github.com/swiftlang/swift-package-manager/blob/main/Sources/Commands/SwiftTestCommand.swift#L1571
+TEST_SHELL="${TEST_SHELL} && ${TEST_CMD} --testing-library swift-testing && [ \$? -eq 0 ] || [ \$? -eq 69 ]"
```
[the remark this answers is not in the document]
marcprux: In my experience, it is [cross-compiled _macros_](https://github.com/swiftlang/swift-package-manager/issues/8094#issuecomment-2455528346) that cause the problem, not the Testing library itself. Testing does use macros, but they are precompiled (unless you are actually building `swift-testing` itself), and have worked fine for every package I've tried, including `swift-png`:

```
marc@zap swift-png % skip android test --copy Sources --copy Tests 

◇ Test run started.
↳ Testing Library Version: 6.3-dev
↳ Target Platform: aarch64-unknown-linux-android
◇ Suite CompressionInternals started.
…
✔ Test EncodeLZ77(_:_:) with 12 test cases passed after 867.646 seconds.
✔ Test EncodePalettes(_:_:) with 18 test cases passed after 868.166 seconds.
✔ Test EncodeiPhoneOptimized(_:_:) with 32 test cases passed after 1084.009 seconds.
✔ Suite Roundtripping passed after 1084.020 seconds.
✔ Test run with 41 tests in 9 suites passed after 1084.024 seconds.
```

I'm happy to put in a logging message warning about this issue, but shouldn't it be for when we identify that the `Package.swift` declares that it uses macros, rather than when any of the tests use Testing?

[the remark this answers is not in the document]
finagolfin: > In my experience, it is [cross-compiled macros](https://github.com/swiftlang/swift-package-manager/issues/8094#issuecomment-2455528346) that cause the problem, not the Testing library itself. Testing does use macros, but they are precompiled (unless you are actually building swift-testing itself)

Yes, of course, that is what I wrote in that issue, but because Testing uses more macros, leaving aside the precompiled ones, I have found it to be a particular issue there.

> have worked fine for every package I've tried, including swift-png

Perhaps I misremembered that one, but the others don't compile.

> I'm happy to put in a logging message warning about this issue, but shouldn't it be for when we identify that the Package.swift declares that it uses macros, rather than when any of the tests use Testing?

You're right: it is best to just mention macros and show the consistent type of error you see, eg `error: Failed opening '/home/finagolfin/swift-foundation/.build/aarch64-unknown-linux-android28/debug/index/store/v5/units/AttributedStringCOWTests.swift.o-2YPXZHICU0J43': No such file or directory`, rather than mention Testing or any specific package.

marcprux:
Could we get some more eyes on this, perhaps by @jakepetroules and/or @justice-adams-apple? I'd like to start testing out various swiftlang projects with Android testing enabled, which will be important for ensuring the quality and stability of the 6.3 release.

.github/workflows/swift_package_test.yml
```diff
@@ -582,45 +586,29 @@ jobs:
           ${{ steps.script_path.outputs.root }}/.github/workflows/scripts/install-and-build-with-sdk.sh --embedded-wasm --flags="$BUILD_FLAGS" ${{ matrix.swift_version }}
 
   android-sdk-build:
-    name: Swift SDK for Android Build (${{ matrix.swift_version }} - ${{ matrix.os_version }} - NDK ${{ matrix.ndk_version }})
+    name: Swift SDK for Android Build (${{ matrix.swift_version }} - NDK ${{ matrix.ndk_version }})
     runs-on: ubuntu-latest
```
justice-adams-apple: Don't we lose some level of coverage here by removing all the SDK builds from docker? Previously we were testing the static SDK builds on a number of different OS's using docker (whatever the repo owner specified with `linux_os_versions`, is there any concern with the loss of coverage for the Swift SDK for Android ? 

Admittedly I'm not Sure how much value that created, maybe that coverage was minimal?

marcprux: I feel like it was overkill, since we are cross-compiling to a different target. It is theoretically possible that some host toolchain issue might get caught by this, but I figured that all the other host toolchains matrix variants that are running in Docker (including the cross-compilation provided by the static SDK) would catch these build issues.

Besides, we can't run the Android emulator tests in Docker due to nested virtualization, so we'd only be able to see if the package builds, not if it passes the tests.

.github/workflows/scripts/android/android-emulator-tests.sh
```diff
@@ -0,0 +1,146 @@
...
+            shift
+            ;;
+    esac
+done
+
+# extract the API level from the end of the triple 
+ANDROID_API="${ANDROID_SDK_TRIPLE/*-unknown-linux-android/}"
+
+# extract the build arch from the beginning of the triple
+ANDROID_EMULATOR_ARCH="${ANDROID_SDK_TRIPLE/-unknown-linux-android*/}"
```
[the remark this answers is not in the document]
marcprux: Added in 8a7fc1c

justice-adams-apple (approved):
we'll definitely want to squash this I assume 👍

shahmishal (approved):
Thanks @marcprux and everyone who provided feedback!

jakepetroules:
This regressed Android builds, causing them to unconditionally use swift-6.3-RELEASE regardless of the `swift_version` parameter.

finagolfin:
You sure? This pull hardly touches `swift_version`.

finagolfin:
[A recent Testing CI run pulls in 6.3.1 automatically](https://github.com/swiftlang/swift-testing/actions/runs/24672184570/job/72346184018?pr=1627#step:10:31) by using this Android workflow.

marcprux:
A more recent failure is that probably related: https://github.com/apple/swift-collections/actions/runs/25028333656/job/73304385415

I think the issue is the `nightly-main` tag, and the host toolchain / Android SDK mismatch causing problems with the build system. I'm iterating on this at https://github.com/swift-android-sdk/github-workflows/pull/3

marcprux:
> [A recent Testing CI run pulls in 6.3.1 automatically](https://github.com/swiftlang/swift-testing/actions/runs/24672184570/job/72346184018?pr=1627#step:10:31) by using this Android workflow.

I _think_ the problem only shows up with the `nightly-main` tag in builds of the Android SDK after 2026-04-27 (as mentioned at https://github.com/swiftlang/github-workflows/pull/269#issue-4347072129), and that successful `nightly-main` [workflow run](https://github.com/swiftlang/swift-testing/actions/runs/24672184570/job/72346183972?pr=1627#step:10:15) was using `swift-DEVELOPMENT-SNAPSHOT-2026-04-01-a`.

finagolfin:
> A more recent failure is that probably related

I mean, that one built fine: the issue is only when uploading the results to the Android emulator.

> I think the problem only shows up with the nightly-main tag in builds of the Android SDK after 2026-04-27

Sounds like a different issue, as Jake was talking about [6.3 release issues above](https://github.com/swiftlang/github-workflows/pull/215#issuecomment-4347092468).

PR #229, opened 2026-01-20
Add CMake build workflow and documentation

FranzBusch:
## Motivation

Swift projects that support CMake as a build system need automated validation to ensure their CMakeLists files stay in sync with the source tree. Without this, CMakeLists files can become stale as source files are added or removed, leading to build failures for users relying on the CMake build path. A reusable workflow helps enforce this validation consistently across Swiftlang projects.

## Modifications

This change introduces a new reusable CMake build workflow that validates and builds Swift projects using CMake. The workflow includes two shell scripts: `cmake-build.sh` which builds projects using CMake and Ninja, and `cmake-update-cmake-lists.sh` which verifies that CMakeLists files accurately reflect the source tree based on a JSON configuration. The workflow runs in a Docker container and accepts inputs for the CMakeLists update configuration, the target directory to build, and the container image to use.

## Result

Projects can now use the cmake_build workflow to automatically verify their CMake build system on every pull request. The workflow ensures CMakeLists files stay up-to-date and that the project builds successfully with CMake, preventing CMake build regressions from being merged.

### DIFF
--- a/.github/workflows/cmake_build.yml
+++ b/.github/workflows/cmake_build.yml
@@ -50,6 +50,8 @@ jobs:
           which curl jq || apt -q update
           which curl || apt -yq install curl
           which jq || apt -yq install jq
+          git config --global --add safe.directory $GITHUB_WORKSPACE
+          cd ${{ inputs.cmake_build_target_directory }}
           CONFIG_JSON='${{ inputs.update_cmake_lists_config }}' FAIL_ON_CHANGES=true ${{ steps.script_path.outputs.root }}/.github/workflows/scripts/cmake-update-cmake-lists.sh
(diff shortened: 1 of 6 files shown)


etcwilde:
When CMake builds break, it's usually gets assigned either to me or @compnerd. I can't speak for him, but having had to help folks debug builds where this script is running, I am not willing to take on that responsibility going forward. If you're going to maintain the CMake configuration for any project using this script, maybe. I'm not comfortable accepting it as an official swiftlang workflow due to the difficulty of fixing things when it breaks them.

This conversation has been brought up before: https://github.com/swiftlang/github-workflows/pull/115

PR #235, opened 2026-02-13
Check if the locally installed Android NDK matches the wanted NDK version

finagolfin:
We were never actually checking `android_ndk_versions` if ANDROID_NDK_HOME was set, as that [environment variable always will be on the github runners](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md).

This caused [build issues when trying to use NDK 28c](https://github.com/swiftlang/swift-testing/actions/runs/21941745673/job/63368987745?pr=1546#step:10:737), as this official workflow was [always linking against the default NDK 27d on the GitHub runners](https://github.com/swiftlang/swift-testing/actions/runs/21941745673/job/63368987745?pr=1546#step:10:213) instead.

@marcprux, guessing this worked locally for you before, but won't on build environments like the GitHub runners.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -52,7 +52,9 @@ jobs:
       # Android
       android_sdk_pre_build_command: |
         cd tests/TestPackage
+      enable_android_sdk_build: true
       enable_android_sdk_checks: true
+      android_ndk_versions: "[\"r28c\"]"
       # Windows
       windows_build_command: |
         cd tests/TestPackage


.github/workflows/scripts/install-and-build-with-sdk.sh
```diff
@@ -653,18 +663,28 @@ install_android_sdk() {
...
-        rm ndk.zip
-        export ANDROID_NDK_HOME="${PWD}"/android-ndk-"${android_ndk_version}"
+        install_android_ndk "$android_ndk_version"
+    elif ! grep -q "Pkg.ReleaseName = ${android_ndk_version}" "${ANDROID_NDK_HOME}/source.properties"; then
+        log "Android NDK $android_ndk_version did not match $(grep "Pkg.ReleaseName" "$ANDROID_NDK_HOME/source.properties")"
+        # Check if the correct NDK is already cached in the same directory, as
+        # it often is on GitHub runners, and use it if so
+        ndk_release=$(echo "$android_ndk_version" | tr -d "[:alpha:]")
+        try_ndk_path=$(find "$(dirname "${ANDROID_NDK_HOME}")" -name "${ndk_release}*" -maxdepth 1)
+        if [[ -d "${try_ndk_path}" ]] && grep -q "Pkg.ReleaseName = ${android_ndk_version}" "${try_ndk_path}/source.properties"; then
```
marcprux: So if I want ndk28c, and it finds a "28*" directory, it will check the `source.properties` file to make sure it is really "ndk28c", and if so, use it. Otherwise (e.g., if it is "ndk28a"), fall back to downloading it as if a matching cache directory did not exist. Correct?

finagolfin: Yep, exactly :+1:

finagolfin:
@justice-adams-apple, mind reviewing?

PR #246, opened 2026-03-19
Remove workaround for bug triggered by Android environment variable

finagolfin:
We had to unset this common Android env var to work around the compiler bug, but it should be fixed in the compiler now.

### DIFF
--- a/.github/workflows/scripts/install-and-build-with-sdk.sh
+++ b/.github/workflows/scripts/install-and-build-with-sdk.sh
@@ -797,11 +797,6 @@ build() {
 
             log "Running: $build_command"
 
-            # clear the ANDROID_NDK_ROOT environment variable if it is set
-            # due to https://github.com/swiftlang/swift-driver/pull/1879
-            # otherwise build error: missing required module 'SwiftAndroid'
-            export ANDROID_NDK_ROOT=""
-
             if eval "$build_command"; then
                 log "✅ Swift build with Android Swift SDK completed successfully"
             else


jakepetroules:
@finagolfin Can we get this merged?

finagolfin:
The last 6.3 snapshot didn't have the env var fix you approved in swift-driver, so I am waiting for the next 6.3 nightly snapshot to be tagged before merging this, else this will still break `nightly-6.3` for Android.

Since unsetting this env var should have no effect on 6.3 and trunk, I'm leaving this workaround in until all three 6.3 and 6.4 builds have the fix.

finagolfin:
Alright, now that a new 6.3 snapshot was tagged, ready to get this in.

finagolfin:
Anyone know why all the PR workflows are no longer being run on this pull?

finagolfin:
@justice-adams-apple, this is ready to go.

finagolfin:
@ahoppen, just need someone to review this small cleanup.

ahoppen:
Do we have any GitHub actions running that might use a Swift version that doesn’t contain the fix for this issue yet?

marcprux:
> Do we have any GitHub actions running that might use a Swift version that doesn’t contain the fix for this issue yet?

I don't think so. The default Android SDK versions for the workflow:

https://github.com/swiftlang/github-workflows/blob/3280fe0808cbace7325c931be850430637ae383a/.github/workflows/swift_package_test.yml#L73-L76

So unless anyone has customized `android_sdk_versions` to explicitly be `6.3.0` (and a quick search against the public repositories doesn't indicate that anyone has), everyone should be running against 6.3.1 or later, and thereby have the fix (https://github.com/swiftlang/swift-driver/pull/1879 and https://github.com/swiftlang/swift-driver/pull/2107) in place.

If any repositories _are_ explicitly using 6.3.0, the canary will be:

```
<unknown>:0: error: missing required module 'SwiftAndroid'
```

finagolfin:
6.3.0 has the fix too: the issue was the last 6.3 snapshot before the release, which is no longer a problem with the newer 6.3 snapshots.

@ahoppen, we don't support anything before 6.3 for Android, so this is fixed in all current branches now.

finagolfin:
@marcprux, see the commit that fixed this, swiftlang/swift-driver@7d6b844f0c2497a997770a11536598b187066be9, the github tag list shows that it made it into both the 6.3 and 6.3.1 releases.

marcprux:
Oh right, I guess I forgot that it got in before 6.3.0. So yes, this should work everywhere.

justice-adams-apple (approved):
thanks!

finagolfin:
@ahoppen, mind merging? Only reviewers have merge rights on this repo.

PR #265, opened 2026-04-17
fix NDK r29 build error

charles-zablit:
https://github.com/swiftlang/swift-tools-support-core/ fail to build because of the [Pull request / Test / Swift SDK for Android Build (nightly-main - NDK r27d) (pull_request)](https://github.com/swiftlang/swift-tools-support-core/actions/runs/24252205555/job/71678817302?pr=555) job.

This is because of a version conflict betwen `ubuntu-latest`'s NDK and swift's one.

### DIFF
--- a/.github/workflows/scripts/install-and-build-with-sdk.sh
+++ b/.github/workflows/scripts/install-and-build-with-sdk.sh
@@ -797,10 +797,17 @@ build() {
 
             log "Running: $build_command"
 
-            # clear the ANDROID_NDK_ROOT environment variable if it is set
-            # due to https://github.com/swiftlang/swift-driver/pull/1879
-            # otherwise build error: missing required module 'SwiftAndroid'
+            # Clear NDK environment variables to prevent tools from picking up a
+            # different NDK version's clang headers, which would conflict with the
+            # Swift SDK's bundled clang resources (module.modulemap redefinition).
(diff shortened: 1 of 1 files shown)


charles-zablit:
cc: @justice-adams-apple (can't add reviewers)

finagolfin:
@jakepetroules, do you see why I don't like using these env vars? It can become a mess, because they are hidden and apply globally to _all_ local development, as opposed to config files that can be applied per-SDK or even per-target-triple. The log message I've been insisting that swift-build put out when choosing the NDK, if there's no config file, would greatly help users [in situations like this](https://github.com/swiftlang/swift-tools-support-core/actions/runs/24252205555/job/71678817302?pr=555#step:10:213).

That said, @charles-zablit, the better solution here may be to only build that Swift package with NDK 28 or 29 with the nightly-main snapshots, as [the trunk CI has switched to building the Android SDK itself with NDK 28c](https://forums.swift.org/t/android-api-minimum-for-the-swift-sdk-for-android/82874/37), and there is no guarantee that other NDKs will work. We plan to switch that trunk SDK CI to the upcoming LTS NDK 30, after which support for NDK 27 will likely be dropped, so might as well get ahead of that now.

finagolfin:
Better to close this, @charles-zablit, as though I've made clear my disdain for these env vars above, 😉 better to not hack around them like this, as I'm also removing the env var workaround we previously added in #246, once the next 6.3 snapshot is tagged.

PR #269, opened 2026-04-29
Change default Android SDK build command

jakepetroules:
Updated default build command for Android SDK to use 'swift build' instead of 'swift build --build-system native'.

The 4/27 snapshot contains the requisite changes for Android to work.

For Wasm, we want to wait until https://github.com/swiftlang/swift-build/pull/1348 lands before unpinning.

### DIFF
--- a/.github/workflows/scripts/install-and-build-with-sdk.sh
+++ b/.github/workflows/scripts/install-and-build-with-sdk.sh
@@ -350,7 +350,6 @@ find_latest_sdk_snapshot() {
 SWIFT_VERSION_BRANCH=""
 ANDROID_SDK_TAG=""
 ANDROID_SDK_CHECKSUM=""
-ANDROID_SDK_PATH_SEP="-"
 
 STATIC_LINUX_SDK_TAG=""
 STATIC_LINUX_SDK_CHECKSUM=""
@@ -621,6 +620,17 @@ if [[ "$INSTALL_ANDROID" == true ]]; then
             exit 0
         fi
     fi
+
+    # Export the resolved Android SDK tag so subsequent workflow steps
+    # (e.g. android-emulator-tests.sh) pick the same SDK we just installed
+    # rather than guessing via 'find ... | tail -n 1'
+    if [[ -n "${GITHUB_ENV:-}" ]]; then
+        {
+            echo "SWIFT_EXECUTABLE_FOR_ANDROID_SDK=${SWIFT_EXECUTABLE_FOR_ANDROID_SDK}"
+            echo "SWIFT_ANDROID_SDK_TAG=${ANDROID_SDK_TAG}"
+            echo "SWIFT_ANDROID_SDK_BUNDLE=${ANDROID_SDK_TAG}_android.artifactbundle"
+        } >> "$GITHUB_ENV"
+    fi
 fi
 
 if [[ "$INSTALL_STATIC_LINUX" == true ]]; then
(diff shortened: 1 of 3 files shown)


marcprux:
I think the fix in https://github.com/swift-android-sdk/github-workflows/pull/3 should resolve the `nightly-main` build errors. Do you want to pull those changes into this PR, or should I file a separate PR against `swiftlang/github-workflows`? (I had filed it against `swift-android-sdk/github-workflows` because I don't have permission to run the workflows in `swiftlang/github-workflows` for testing).

jakepetroules:
Thanks, I incorporated your changes.

marcprux:
It looks like it is still [failing](https://github.com/swiftlang/github-workflows/actions/runs/25194053598/job/73870509566), but the [new log message](https://github.com/swiftlang/github-workflows/actions/runs/25194053598/job/73870509566#step:10:291) indicates that it is using the right host toolchain:

```
** Host toolchain for Android build: /home/runner/.swift-toolchains/swift-DEVELOPMENT-SNAPSHOT-2026-04-28-a/usr/bin/swift
```

That new failure seems to be coming from:

```
find: ‘/home/runner/work/github-workflows/github-workflows/tests/TestPackage/.build/x86_64-unknown-linux-android28/debug’: No such file or directory
```

Should we be expecting a different output folder?

marcprux:
I think the issue here is that `swift-build` changed where they are outputting the tests, which is going to break the assumptions we make in the Android testing support here, as well as in skip (https://github.com/skiptools/skipstone/issues/244) and the `swift-android-action` GitHub workflow (https://github.com/skiptools/swift-android-action/issues/21).

So for example, we need to change the assumption that the test case is built at `.build/aarch64-unknown-linux-android28/debug/swift-algorithmsPackageTests.xctest` and instead copy it from `.build/out/Products/Debug-android/SwiftAlgorithmsTests-test-runner`.

.github/workflows/swift_package_test.yml
```diff
@@ -171,8 +171,7 @@ on:
       android_sdk_build_command:
         type: string
         description: "Command to use when building the package with the Swift SDK for Android"
-        # Temporarily use native build system on Android due to https://github.com/swiftlang/swift/issues/88282
-        default: "swift build --build-system native"
+        default: "swift build"
```
marcprux, line 174: ❤️

.github/workflows/scripts/install-and-build-with-sdk.sh
```diff
@@ -717,7 +733,11 @@ install_android_sdk() {
         fi
     fi
 
-    ./swift-sdks/"${android_sdk_bundle_name}"/swift-android/scripts/setup-android-sdk.sh
+    # run the bundled setup script with the matched toolchain in front of PATH
+    # so any 'swift' it invokes resolves to the version that matches the SDK
+    local toolchain_bin_dir
+    toolchain_bin_dir="$(dirname "$SWIFT_EXECUTABLE_FOR_ANDROID_SDK")"
+    PATH="${toolchain_bin_dir}:${PATH}" ./swift-sdks/"${android_sdk_bundle_name}"/swift-android/scripts/setup-android-sdk.sh
```
marcprux: The dream is that we will eventually no longer require this script at all, right? In my equivalent actions at https://github.com/skiptools/swift-android-action/blob/8db781febb546110d48dee035131db8f4e259e91/action.yml#L463-L469, I check whether the script exists, and if not, then I don't bother executing it, so that when it gets removed from the SDK distribution we will continue to work. You may want to add a similar check here.

marcprux (approved):

.github/workflows/scripts/install-and-build-with-sdk.sh
justice-adams-apple: Might be worth future proofing depending on if we know when that script will go away

justice-adams-apple (approved):

.github/workflows/scripts/install-and-build-with-sdk.sh
jakepetroules: Done!

PR #233, opened 2026-01-28
Add PR Dependency check workflow

bkhouri:
While working on a project, I tend to build one PR on top of another.    And although The later PR can be merged and include all the child PR, it's not ideal as it can lead to the "parent" PR being closed without having any references to approval - as the approvals were given in a child PR.

Sometimes SwiftPM PR's are depend on a change in https://github.com/swiftlang/swift-build, or even changes in https://github.com/swiftlang/swift

Here's are examples of the two scenarios:
- https://github.com/swiftlang/swift-package-manager/pull/9039
- https://github.com/swiftlang/swift-package-manager/pull/9417
- https://github.com/swiftlang/swift-package-manager/pull/8092

PR #270, opened 2026-04-29
extract PR Dependency Check for github-workflows into its own file

justice-adams-apple:
This is a fix for https://github.com/swiftlang/github-workflows/pull/233/

Because the Pr dependency check has different triggers, extract it into it's own file for this repository that way we can use the proper triggering syntax when PR's are opened


FranzBusch (approved):

shahmishal:
Thanks @justice-adams-apple

shahmishal (approved):

PR #275, opened 2026-05-22
[freebsd] cd into the proper dir to run package test

justice-adams-apple:
Fix the workflow action 

```
  Swift version 6.3-dev (LLVM 972b62858355d07, Swift 3f8c798cfe25bbb)
  Target: x86_64-unknown-freebsd14.3
  Build config: +assertions
  error: Could not find Package.swift in this directory or any of its parent directories.
```

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -63,6 +63,8 @@ jobs:
       windows_os_versions: '["windows-2022", "windows-11-arm"]'
       # FreeBSD
       enable_freebsd_checks: true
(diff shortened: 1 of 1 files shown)


FranzBusch (approved):

shahmishal (approved):

jakepetroules:
Thanks! Previously I assumed I missed something in swift_package_test (since we have the conditions like `if [ "${{ github.repository }}" = "swiftlang/github-workflows" ]; then`) but completely missed that this was expected for all the other platforms as well.

PR #280, opened 2026-06-07
Fix Android build for non-default NDKs with new swift-build build system, by updating ANDROID_NDK_ROOT too

finagolfin:
swiftlang/swift-testing#1745 showed that the updated NDK env var isn't being read on CI, see if it is set.

### DIFF
--- a/.github/workflows/scripts/install-and-build-with-sdk.sh
+++ b/.github/workflows/scripts/install-and-build-with-sdk.sh
@@ -815,12 +815,16 @@ build() {
 
         alias swift='$SWIFT_EXECUTABLE_FOR_ANDROID_SDK'
 
+        log "Using NDK at $ANDROID_NDK_HOME"
+
         # This can become a single invocation in the future when `swift build` supports multiple Android triples at once
         for android_sdk_triple in "${ANDROID_SDK_TRIPLES[@]}" ; do
             if [[ "$SWIFT_VERSION_INPUT" == "6.3" || "$SWIFT_VERSION_INPUT" == "nightly-6.3" ]]; then
                 local build_command="$SWIFT_BUILD_COMMAND --swift-sdk ${android_sdk_triple}"
             else
                 local build_command="$SWIFT_BUILD_COMMAND --swift-sdk ${sdk_name} --triple ${android_sdk_triple}"
+                # Work around swift-build issue with ANDROID_NDK_ROOT overriding ANDROID_NDK_HOME
+                export ANDROID_NDK_ROOT="${ANDROID_NDK_HOME}"
             fi
             if [[ -n "$SWIFT_BUILD_FLAGS" ]]; then
                 build_command="$build_command $SWIFT_BUILD_FLAGS"


.github/workflows/scripts/install-and-build-with-sdk.sh
```diff
@@ -814,7 +814,10 @@ build() {
         local sdk_name="${ANDROID_SDK_TAG}_android"
 
         alias swift='$SWIFT_EXECUTABLE_FOR_ANDROID_SDK'
-
+    
+        log "Using NDK at $ANDROID_NDK_HOME"
```
finagolfin, line 818: @jakepetroules, `swift-build` is not using this environment variable properly on the GitHub CI, where it is [correctly set to 28c](https://github.com/swiftlang/github-workflows/actions/runs/27092508867/job/79958541127#step:9:312) before the build starts, but [the internal compile shows it using NDK 27 instead](https://github.com/swiftlang/github-workflows/actions/runs/27092508867/job/79958541127#step:9:531). I tried setting it in `$GITHUB_ENV` also in the next line, but that made no difference.

This is why I pushed to not have these flaky env vars be the primary way we configure the NDK for SwiftPM, and to always notify the user that the env var is being used and with which NDK path.

Let me know if you have any idea about this bug.

jakepetroules: Swift Build checks both ANDROID_NDK_ROOT and ANDROID_NDK_HOME, with the first taking precedence. Since GitHub Actions sets both, _ROOT is pointing to 27 still.

Go ahead and update GitHub actions to set both variables.

It's probably a good idea to emit a diagnostic in Swift Build if those two are set to different values.

finagolfin: Yep, that was it, will clean this up and submit as a workaround for now.

finagolfin:
@justice-adams-apple, Android fix I'd like to get in once Jake signs off.

finagolfin:
Ping @jakepetroules, easy review.

.github/workflows/scripts/install-and-build-with-sdk.sh
```diff
@@ -815,12 +815,16 @@ build() {
...
+
         # This can become a single invocation in the future when `swift build` supports multiple Android triples at once
         for android_sdk_triple in "${ANDROID_SDK_TRIPLES[@]}" ; do
             if [[ "$SWIFT_VERSION_INPUT" == "6.3" || "$SWIFT_VERSION_INPUT" == "nightly-6.3" ]]; then
                 local build_command="$SWIFT_BUILD_COMMAND --swift-sdk ${android_sdk_triple}"
             else
                 local build_command="$SWIFT_BUILD_COMMAND --swift-sdk ${sdk_name} --triple ${android_sdk_triple}"
+                # Work around swift-build issue with ANDROID_NDK_ROOT overriding ANDROID_NDK_HOME
+                export ANDROID_NDK_ROOT="${ANDROID_NDK_HOME}"
```
jakepetroules, line 827: Can you please put this closer to line 667 so both vars are in the same place?

jakepetroules (changes requested):

.github/workflows/scripts/install-and-build-with-sdk.sh
finagolfin: I put it here so it isn't messed with for 6.3 builds. I'd like to get rid of this instead for `swift sdk configure --sdk-root-path`, once that is working for Android, so I see this as a temporary fix anyway.

jakepetroules: I think it's fine, there's no reason it needs to be excluded for 6.3 builds.

finagolfin: You're probably right, but given stuff like swift-driver incorrectly invoking it before 6.3, I played it safe.

finagolfin:
@justice-adams-apple, I'd like to get this in now, else the snapshot builds for Android are broken for other NDKs. I will be modifying this same line to use SwiftPM instead of env vars in the coming weeks, once that `swift sdk configure` feature is enabled for Android again in `swift-build`, so this location is fine.

justice-adams-apple (approved):

PR #282, opened 2026-06-09
Docs: provide option to skip analysis in doc-check

### LINKED ISSUE #281: `ADDITIONAL_DOCC_ARGUMENTS: unbound variable` when running the Documentation check workflow
The `.github/workflows/scripts/check-docs.sh` script generated an unbound variable failure.  Rightfully so, the variable is never defined.  In addition, the repository does not have any documentation explaining how, or whether, the variable should be defined by the caller.

Steps to reproduce:
1. Clone this repository under the directory `GitHub-workflows`
2. Clone a Swift PM fork as a sibling to the GitHub workflow repository 
   ```
    git clone -b t/6.4.x/gh10186_gha_update-doc-check-to-check-target https://github.com/bkhouri/swift-package-manager.git
    ```
3. Change directory to the SwiftPM repo and run the following:
    ```
    cd swift-package-manager
    ../github-workflows/.github/workflows/scripts/check-docs.sh
    ```

**Actual**

```
❯ ../github-workflows/.github/workflows/scripts/check-docs.sh
** swift-docc-plugin already exists
** Checking documentation targets...
** Checking target PackageManagerDocs...
../github-workflows/.github/workflows/scripts/check-docs.sh: line 59: ADDITIONAL_DOCC_ARGUMENTS: unbound variable
```

bkhouri:
Some projects may not want to run the `--analyze` option on the DocC generation until they are ready.

Provide an workflow input option to skip the analyze.  While at it, update the `check-docs.sh` script to accept arugments instead of relying on environment variables being set.  This allows more easily run the script at-desk.

Fixes #281

### DIFF
--- a/.github/workflows/soundness.yml
+++ b/.github/workflows/soundness.yml
@@ -31,6 +31,14 @@ on:
         type: string
         description: "Additional arguments that should be passed to docc"
         default: ""
+      docs_check_targets:
+        type: string
+        description: "List of targets to check for documentation."
+        default: ""
+      docs_check_analyze:
+        type: boolean
+        description: "Boolean to pass --analyze to the docs check. Defaults to true."
+        default: true
       docs_check_macos_enabled:
         type: boolean
         description: "Boolean to enable the macOS docs check job. Defaults to false."
@@ -51,6 +59,14 @@ on:
         type: string
         description: "Additional arguments that should be passed to docc for the macOS docs check job."
         default: ""
+      docs_check_macos_targets:
+        type: string
+        description: "List of targets to check for documentation."
+        default: ""
+      docs_check_macos_analyze:
+        type: boolean
+        description: "Boolean to pass --analyze to the macOS docs check. Defaults to true."
+        default: true
       unacceptable_language_check_enabled:
         type: boolean
         description: "Boolean to enable the acceptable language check job. Defaults to true."
@@ -99,13 +115,17 @@ on:
         type: string
         description: "Linux command to execute before building the Swift package"
         default: ""
+      concurrency_group_suffix:
+        type: string
+        description: "Optional suffix appended to the concurrency group. Use this when calling soundness.yml from a matrix so that matrix entries don't cancel each other."
+        default: ""
 
 permissions:
   contents: read
 
 ## We are cancelling previously triggered workflow runs
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}-soundness
+  group: ${{ github.workflow }}-${{ github.ref }}-soundness-${{ inputs.concurrency_group_suffix }}
   cancel-in-progress: true
 
 jobs:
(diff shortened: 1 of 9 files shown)


bkhouri:
Since I am unable to re-trigger the failed GitHub Actions, I'm closing and re-opening the PR to trigger them

docs/soundness-docs-check.md
```diff
@@ -0,0 +1,149 @@
...
+Running both lets you catch documentation issues that only surface on one toolchain.
+
+Documentation warnings (and, by default, DocC analyzer findings) cause the job to fail.
+
+## Requirements
+
+For either job to run, your repository needs:
+
+1. A `Package.swift` (or any `Package*.swift`) at the repository root.
+2. A `.spi.yml` at the repository root listing the targets to document. Read the [official documentation](https://swiftpackageindex.com/SwiftPackageIndex/SPIManifest/1.12.0/documentation/spimanifest/commonusecases) on how to generate the `.spi.yml`. For example:
```
bkhouri: issue (blocking): this is no longer the case.  update the documentation.

.github/workflows/soundness.yml
```diff
@@ -31,6 +31,14 @@ on:
         type: string
         description: "Additional arguments that should be passed to docc"
         default: ""
+      docs_check_targets:
+        type: string
+        description: "List of targets to check for documentation. Defaults to empty string."
```
FranzBusch: Empty string means that it looks at `.spi.yml` right?

.github/workflows/soundness.yml
```diff
@@ -51,6 +59,14 @@ on:
         type: string
         description: "Additional arguments that should be passed to docc for the macOS docs check job."
         default: ""
+      docs_check_macos_targets:
+        type: string
+        description: "List of targets to check for documentation. Defaults to empty string."
```
FranzBusch: Same here

docs/soundness-docs-check.md
```diff
@@ -0,0 +1,162 @@
...
+
+## Requirements
+
+### For both jobs
+
+1. **A `Package.swift`** (or any `Package*.swift`) at the repository root.
+
+2. **At least one documentation target**, supplied via either:
+   - the `docs_check_targets` / `docs_check_macos_targets` input, **or**
+   - a `documentation_targets` entry in a `.spi.yml` file at the repository root.
```
FranzBusch: Can we swap this since I expect the second one to be the one we want folks to use

docs/soundness-docs-check.md
```diff
@@ -0,0 +1,162 @@
...
+
+   If neither is provided, the job exits successfully without checking anything.
```
FranzBusch, line 22: This might be a bit unexpected. Shouldn't this fail the job?

FranzBusch (commented):
Overall LGTM a few minor nits. Thanks for adding a doc for the check!

.github/workflows/soundness.yml
bkhouri: Empty should be the current behaviour :)

docs/soundness-docs-check.md
bkhouri: I opted to keep the current behaviour, which does not fail the job if `.spi.yml` is not defined.

docs/soundness-docs-check.md
bkhouri: I swapped them.

FranzBusch (approved):

PR #290, opened 2026-06-25
Update dock check syntax issue

### LINKED ISSUE #291: docs-check broken by #282: quoted doc target + docs_check_targets input ignored in containers
The `docs-check` soundness job fails when the documentation target is read from `.spi.yml`:

```
error: no target named '"AWSLambdaRuntime"'
compatible targets: 'AWSLambdaRuntime', ...
```

The target name reaches `swift package plugin generate-documentation` wrapped in literal double-quotes. Two separate bugs, both introduced in #282, cause this — and they compound, so fixing only one is not enough.

## Bug 1 — `check-docs.sh` drops the `yq -r` flag

`.github/workflows/scripts/check-docs.sh` reads targets from `.spi.yml` without `-r`:

```bash
docs_targets=$(yq ".builder.configs[] | select(.documentation_targets[] != \"\") | .documentation_targets[]" .spi.yml)
```

The Swift Ubuntu containers `apt install` the Python `yq` (a `jq` wrapper). Without `-r` it prints JSON-quoted strings, so `$target` becomes `"AWSLambdaRuntime"` (quotes included) and is passed verbatim to `--target`.

Verified in `swift:6.3-noble`:

```console
$ yq ".builder.configs[] | .documentation_targets[]" .spi.yml
"AWSLambdaRuntime"      # without -r
$ yq -r ".builder.configs[] | .documentation_targets[]" .spi.yml
AWSLambdaRuntime        # with -r
```

The pre-#282 script used `yq -r` and worked. (Note line 106 of the same script still correctly uses `-r`.)

**Fix:** restore `-r` on the extraction line.

## Bug 2 — `soundness.yml` uses `[[` under `sh`

The intended workaround is to pass targets explicitly via the `docs_check_targets` input, which makes the script skip the `.spi.yml` path. But the `Run documentation check` step in `soundness.yml` builds that argument with bash-only `[[`:

```yaml
run: |
  doc_target_arg=""
  if [[ "${DOCS_TARGETS}" != "" ]] ; then
    doc_target_arg="--doc-targets ${DOCS_TARGETS}"
  fi
  ...
```

The `docs-check` job runs in a container, where the default step shell is `sh` (dash), and the step has no `shell: bash`. Under dash, `[[` is not a builtin:

```
/__w/_temp/....sh: 3: [[: not found
```

So `doc_target_arg` stays empty, `--doc-targets` is never passed, and the script falls back to the broken path in Bug 1. This means the `docs_check_targets` workaround silently has no effect.

**Fix:** add `shell: bash` to the documentation-check step(s), or replace `[[ ... ]]` with POSIX `[ ... ]`.

## Workaround for consumers

Until both are fixed, pin to the commit before #282, which has neither bug:

```yaml
uses: swiftlang/github-workflows/.github/workflows/soundness.yml@9abcbf2
```

## Affected

`main` of swiftlang/github-workflows, since #282.

bkhouri:
Fixes: #291

### DIFF
--- a/.github/workflows/soundness.yml
+++ b/.github/workflows/soundness.yml
@@ -18,7 +18,7 @@ on:
       api_breakage_check_container_image:
         type: string
         description: "Container image for the API breakage check job. Defaults to latest Swift Ubuntu image."
-        default: "swift:6.2-noble"
+        default: "swift:6.3-noble"
       docs_check_enabled:
         type: boolean
         description: "Boolean to enable the docs check job. Defaults to true."
@@ -26,7 +26,7 @@ on:
       docs_check_container_image:
         type: string
         description: "Container image for the docs check job. Defaults to latest Swift Ubuntu image."
-        default: "swift:6.2-noble"
+        default: "swift:6.3-noble"
       docs_check_additional_arguments:
         type: string
         description: "Additional arguments that should be passed to docc"
@@ -94,7 +94,7 @@ on:
       format_check_container_image:
         type: string
         description: "Container image for the format check job. Defaults to latest Swift Ubuntu image."
-        default: "swift:6.2-noble"
+        default: "swift:6.3-noble"
       shell_check_enabled:
         type: boolean
         description: "Boolean to enable the shell check job. Defaults to true."
@@ -200,10 +200,6 @@ jobs:
           else
             echo "root=$GITHUB_WORKSPACE/github-workflows" >> $GITHUB_OUTPUT
           fi
-      - name: Pre-build
-        if: ${{ inputs.linux_pre_build_command }}
-        # zizmor: ignore[template-injection]
-        run: ${{ inputs.linux_pre_build_command }}
       - name: Run documentation check
         env:
           ADDITIONAL_DOCC_ARGUMENTS: ${{ inputs.docs_check_additional_arguments }}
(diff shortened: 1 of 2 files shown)


FranzBusch (approved):

.github/workflows/soundness.yml
```diff
@@ -200,25 +200,27 @@ jobs:
           else
             echo "root=$GITHUB_WORKSPACE/github-workflows" >> $GITHUB_OUTPUT
           fi
-      - name: Pre-build
-        if: ${{ inputs.linux_pre_build_command }}
-        # zizmor: ignore[template-injection]
-        run: ${{ inputs.linux_pre_build_command }}
+      # - name: Pre-build
```
FranzBusch: Should this stay commented out?

FranzBusch (changes requested):
Can we add a test for this so we don't regress this again?

bkhouri:
> Can we add a test for this so we don't regress this again?

The is a test in the Workflow.  It was previously giving a false positive :(

FranzBusch (approved):

.github/workflows/soundness.yml
```diff
@@ -200,23 +200,21 @@ jobs:
           else
             echo "root=$GITHUB_WORKSPACE/github-workflows" >> $GITHUB_OUTPUT
           fi
-      - name: Pre-build
-        if: ${{ inputs.linux_pre_build_command }}
-        # zizmor: ignore[template-injection]
-        run: ${{ inputs.linux_pre_build_command }}
```
FranzBusch, line 206: Why have we moved this inline?

bkhouri: Great question and I'm glad someone asked.  The `GitHub-workflows` PR builds call this soundness check and set a `linux_pre_build_command` as following:

https://github.com/swiftlang/github-workflows/blob/d65527ae024104f80c3c9a204e2255ff6e6dbf24/.github/workflows/pull_request.yml#L117-L119

We essentially do a `cd tests/TestPackage` to change directory to the Sample package.  However, when `Run documentation check` step was executed, it was against the repository root.  This was one solution I found that address the use case, but I also considered adding a new input parameter with `docs_check_package_root_directory`, that would specify a relative directory to the SwiftPM Package to build  for the docs.

FranzBusch (approved):

PR #292, opened 2026-06-26
Add a build verdict job

bkhouri:
Add a build verdict job that will fail when any dependent workflow fails. The workflow also emits, on the console, all required job names that failed.

Depends on #290


FranzBusch (commented):
So this is only for this repo. I would like us to discuss and agree if this is expected from all repos that depend on this one as a guidance.

PR #293, opened 2026-06-26
Update default Xcode versions in workflow to Xcode 27 Beta 2

shahmishal:
Added Xcode version '27.b2' to the default list.

### DIFF
--- a/.github/workflows/swift_package_test.yml
+++ b/.github/workflows/swift_package_test.yml
@@ -9,7 +9,7 @@ on:
       macos_xcode_versions:
         type: string
         description: "Xcode version list (JSON)"
-        default: "[\"16.4\", \"26.3\", \"26.4\"]"
+        default: "[\"16.4\", \"26.3\", \"26.4\", \"27.b2\"]"
       macos_exclude_xcode_versions:
         type: string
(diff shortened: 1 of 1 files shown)


.github/workflows/swift_package_test.yml
```diff
@@ -9,7 +9,7 @@ on:
       macos_xcode_versions:
         type: string
         description: "Xcode version list (JSON)"
-        default: "[\"16.4\", \"26.3\", \"26.4\"]"
+        default: "[\"16.4\", \"26.3\", \"26.4\", \"27.b2\"]"
```
madrob, line 12: Could this use the "latest" symlink instead so that we don't have to update every time the Xcode beta version changes?

shahmishal: We need to use a specific version so that people can exclude it and clearly understand what they are getting with a specific tag.

madrob: So that means we need to have at least two beta versions on each runner? What are the update window guarantees that we must provide? How many versions at a time will the runner need to carry so that consumers of this have enough time to update their workflows?

PR #301, opened 2026-07-08
Add macOS nightly build support via Swiftly

### LINKED ISSUE #240: Add support for running builds using macOS OSS Toolchain
As a software developer contributing to Swift project, I want to be able to run GitHub Actions using the macOS OSS toolchain so I can verify my changes against a macOS toolchain that contains my changes.

--

The current macOS GitHub Actions run against a pre-installed Xcode.  This has a released version of Swift toolchain.  There are time where the changes I want to validate are in an unreleased Swift toolchain version (e.g.: `main nightly`), but there is currently no way to execute GitHub Actions against an unreleased version of the Swift Toolchain. 

This issue is to track adding GitHub Actions support for running against a macOS OSS Toolchain. 

One use case: Swift For Embedded does not require Xcode.

shahmishal:
Adds a new opt-in macos-swiftly-build job that pairs an Xcode version with a Swiftly-managed Swift toolchain (defaulting to main-snapshot), since xcrun cannot see toolchains that Swiftly selects on PATH.

Fixes #240

### DIFF
--- a/.github/workflows/swift_package_test.yml
+++ b/.github/workflows/swift_package_test.yml
@@ -312,7 +312,7 @@ jobs:
               printf "%s\n" $i >> $GITHUB_ENV
             done
       - name: Select Xcode
-        run: echo "DEVELOPER_DIR=/Applications/Xcode_${{ matrix.xcode_version }}.app" >> $GITHUB_ENV
+        run: echo "DEVELOPER_DIR=/Applications/Xcode_${{ matrix.xcode_version }}.app/Contents/Developer" >> $GITHUB_ENV
       - name: Swift version
         run: xcrun swift --version
       - name: Clang version
@@ -350,7 +350,7 @@ jobs:
               printf "%s\n" $i >> $GITHUB_ENV
             done
       - name: Select Xcode
-        run: echo "DEVELOPER_DIR=/Applications/Xcode_${{ matrix.xcode_version }}.app" >> $GITHUB_ENV
+        run: echo "DEVELOPER_DIR=/Applications/Xcode_${{ matrix.xcode_version }}.app/Contents/Developer" >> $GITHUB_ENV
       - name: Set up Swiftly environment
         run: |
           . "$HOME/.swiftly/env.sh"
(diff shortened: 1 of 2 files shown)


.github/workflows/swift_package_test.yml
```diff
@@ -139,6 +143,10 @@ on:
         type: string
         description: "macOS command to build and test the package"
         default: "xcrun swift test"
+      macos_swiftly_build_command:
+        type: string
+        description: "macOS command to build and test the package when using a Swiftly-managed toolchain (run via `swiftly run`, so it should not use `xcrun`)"
+        default: "swift test"
```
justice-adams-apple: Should the default be 
```
swiftly run swift build
```
?

shahmishal: Yes, good catch!

justice-adams-apple (approved):

finagolfin:
@marcprux and @kateinoigakukun, wdyt about adding an option to build using the SDK bundles on macOS, now that OSS macOS snapshot toolchains are installed? It could be a boolean like `enable_macos_sdk_builds` and then whatever SDK bundles are enabled, whether Android or wasm or static linux currently enabled separately, would be installed and used to cross-compile the code on macOS too. 👍

marcprux:
> wdyt about adding an option to build using the SDK bundles on macOS, now that OSS macOS snapshot toolchains are installed?

I think it would be worth trying. I haven't tried building the Android SDK on macOS in a while, but the last time I tried there were some non-trivial errors I encountered (mostly related to cross-compilation from Darwin hosts, IIRC).

finagolfin:
> I haven't tried building the Android SDK on macOS in a while, but the last time I tried there were some non-trivial errors I encountered (mostly related to cross-compilation from Darwin hosts, IIRC).

Not what I'm talking about: all our prebuilt SDK bundles work well with the OSS macOS toolchain to cross-compile for linux/Android/wasm. We can provide mac->Android cross-compilation workflows for arbitrary Swift packages now, for those who want to test that and to make sure our prebuilt official SDK bundles keep working well on macOS.

PR #308, opened 2026-08-03
remove amazonlinux2 checkouts

madrob:
AmazonLinux2 was removed from testing in #298 we should also remove the conditional use of checkout@v1 action.

### DIFF
--- a/.github/workflows/swift_package_test.yml
+++ b/.github/workflows/swift_package_test.yml
@@ -447,23 +447,12 @@ jobs:
         run: clang --version
       - name: Checkout repository
         uses: actions/checkout@v7
-        if: ${{ matrix.os_version != 'amazonlinux2' }}
-      - name: Checkout repository
-        uses: actions/checkout@v1
-        if: ${{ matrix.os_version == 'amazonlinux2' }}
       - name: Checkout swiftlang/github-workflows repository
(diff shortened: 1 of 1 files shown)


shahmishal:
We have not removed it from checkout support yet because repositories might still test Amazon Linux 2 with older Swift releases.

madrob:
What's that going to be gated on? https://www.swift.org/platform-support/ doesn't distinguish by Swift version.

shahmishal:
> What's that going to be gated on? https://www.swift.org/platform-support/ doesn't distinguish by Swift version.

https://github.com/swiftlang/project-operations/blob/main/github-policies.md
- Packages **should** support the **latest patch releases** of the **last three minor** official Swift releases.
    - For example, if the currently latest Swift release is `6.2.1` then a package would be expected to support `6.2.1`, `6.1.3` and `6.0.3`.
    - As an exception, new packages are free to start with only supporting the latest official Swift release.

madrob:
That answers half the question, but not "which versions of Swift need to run on AmazonLinux 2?"

I think it's everything `<= 6.3`, but I don't see that written. So the implication is that we would be able to merge this PR after Swift 6.6 is released?

shahmishal:
After next three releases (it might be 6.6 or 7.x) would be correct.

PR #309, opened 2026-08-18
swift_package_test: Switch to using Swift version-based symlinks

lorentey:
This lets workflows pick up the latest version of Xcode that is available on the runners automatically, with minimal effort on the part of package maintainers.

However, any package repository that has its own Xcode version lists (`macos_xcode_versions`, `macos_exclude_xcode_versions`) will need to be updated accordingly, as part of updating to a new github-workflows release with this change. (The exlusion list in particular will fail to function until this is done.)

At the time of writing, version specs in columns 2 or 3 below need to be replaced with the symbolic `swift_6.x` name in column one:

|  Symbolic name | Xcode version spec | Previous releases          |
|----------------|--------------------|----------------------------|
| swift_6.1      | 16.4               | --                         |
| swift_6.2      | 26.3               | 26.0, 26.1, 26.2           |
| swift_6.3      | 26.6               | 26.4, 26.4.1, 26.5         |
| swift_6.4      | 27.b5              | 27.b1, 27.b2, 27.b3, 27.b4 |


FranzBusch (approved):

shahmishal (approved):
Thanks @lorentey!

PR #310, opened 2026-08-18
Add install cmake action

etcwilde:
Adding an action to install a specific version of CMake. The version is verified against a hash before being unpacked and installed.

I have verified this with several versions of CMake on Linux.
I haven't had a setup with macOS or Windows to test against yet, but the GitHub actions docs say that bash is available on [all platforms](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstepsshell) (as is powershell, but I don't want to write powershell) with the note

> When specifying a bash shell on Windows, the bash shell included with Git for Windows is used.

The link format is consistent, with `cmake-<version>-<arch>.(tar.gz|.zip)`. There are OS-specific installer-based packages, but the tarballs and zips are easier to unpack and verify reliably.


.github/actions/install-cmake/readme.md
```diff
@@ -0,0 +1,47 @@
...
+```yaml
+...
+
+steps:
+
+  ...
+
+  - name: Install CMake
+    id: install-cmake
+    uses: swiftlang/github-workflows/.github/workflows/soundness.yml@<version>
```
jmschonfeld: Is this the right path?

etcwilde: No it is not. Nice catch.

etcwilde:
```
-#!/bin/bash
+#!/usr/bin/env bash
```

... really? That's not even part of the license.

FranzBusch:
Where do you see this going? I opened a PR a few months ago to add Cmake based workflows which we are using in the Swift Server world https://github.com/swiftlang/github-workflows/pull/229? Curious to hear your thoughts how this PR might relate to my PR.

etcwilde:
> Where do you see this going? I opened a PR a few months ago to add Cmake based workflows which we are using in the Swift Server world https://github.com/swiftlang/github-workflows/pull/229? Curious to hear your thoughts how this PR might relate to my PR.

I see this allowing folks to install specific version(s) of CMake and then calling it. It's meant to compose with existing workflows by exposing the cmake binary as an output of the action.

e.g.
 - https://github.com/swiftlang/swift-cmake-examples/blob/main/.github/workflows/pull_request.yml
 - https://github.com/swiftlang/swift-foundation/blob/d695107ad2e69e189a6e16cc32d3a05fd251674d/.github/workflows/pull_request.yml#L44

The referenced PR only appears to work on apt-based Linux distros, rewrites the CMakeLists.txt, and then tries to run the build. To update that PR with this action in place, you could drop the `which cmake || apt -yq install cmake` portion and use it instead.

```yaml
      - name: CMake build
        run: |
          which curl cmake ninja || apt -q update
          which curl || apt -yq install curl
          which cmake || apt -yq install cmake
          which ninja || apt -yq install ninja-build
          TARGET_DIRECTORY="${{ 
```

Then reference the output of the action with something like

```yaml
 - name: Run CMake
     shell: bash
     env:
         CMAKE: ${{ steps.install-cmake.outputs.ctest }}
     run: |
       "$CMAKE" -G Ninja -B build -S .
...
```

.github/actions/install-cmake/install-cmake.sh
```diff
@@ -0,0 +1,102 @@
+#!/bin/bash
+##===----------------------------------------------------------------------===##
+##
+## This source file is part of the Swift.org open source project
+##
+## Copyright (c) YEARS Apple Inc. and the Swift project authors
```
shahmishal:
```suggestion
## Copyright (c) 2026 Apple Inc. and the Swift project authors
```

shahmishal:
Please add a workflow to test this part of PR testing.

etcwilde:
@shahmishal and I talked offline about also adding a default workflow that projects can use directly. I will put that up in a separate PR. I'd like to unblock Swift Format and Swift Foundation who are blocked after bumping Swift Syntax to requiring CMake 3.29.

.github/actions/install-cmake/install-cmake.sh
```diff
@@ -0,0 +1,102 @@
...
+    exit 1
+    ;;
+esac
+
+ARCHIVE_URL="https://github.com/Kitware/CMake/releases/download/v${VERSION}/${ARCHIVE}"
+ARCHIVE_PATH="${TEMP_DIR}/${ARCHIVE}"
+
+# 3. Download
+echo "==> Downloading ${ARCHIVE_URL}..."
+curl -fsSL --retry 3 -o "${ARCHIVE_PATH}" "${ARCHIVE_URL}"
```
etcwilde, line 67: For the future and digging through the action marketplace, migrating this script to Javascript should let us drop all of the dependencies since nodejs is supposed to always be available.

shahmishal (approved):

PR #313, opened 2026-08-21
Adding Reusable CMake build workflow

etcwilde:
This workflow absorbs the boilerplate in installing and setting CMake builds across macOS, Linux, and Windows.

The workflow defaults to using CMake 4.0.3, and has the appropriate hashes for each architecture and platform. If you override the version, you will also need to provide the appropriate hashes for the platforms.

Each platform can be optionally enabled or disabled with the `enable_*_checks`. By default, only Linux is tested, but macOS and Windows can both be enabled.

Controlling each build is done with three commands for each platform:

 1. `*_pre_build_command`
 2. `*_configure_command`
 3. `*_build_command`

The pre-build command runs first, allowing you to install dependencies like Ninja. The `configure` command is used to run CMake to configure the build. By default, this runs the platform-equivalent of `$CMAKE -G Ninja -B build -S .`. The build-command is for running the build. By default, it runs the platform-equivalent of `$CMAKE --build build`

Each of these commands has a `CMAKE` and `CTEST` environment variable defined, specifying the path to the installed CMake and CTest binaries.

Linux and Windows both run x86_64 and Arm64, while macOS only builds Arm64.


etcwilde:
@shahmishal, the CMake build workflow you asked for.

shahmishal (approved):

PR #315, opened 2026-09-10
Use the new LTS Android NDK alone for both testing packages and for workflow pulls here

finagolfin:


finagolfin:
[LTS NDK 30 is now the only supported one](https://groups.google.com/g/android-ndk-announce/c/P3UWFg0jdpE), but keep the older ones around until we ship a Swift 6.4 SDK built with NDK 30 and [the GitHub CI have installed it in their runners](https://github.com/actions/runner-images/blob/511e65ce908f72f78db9bb4052d642a8728681cb/images/ubuntu/Ubuntu2404-Readme.md#android).

@marcprux and @justice-adams-apple, please review.

marcprux (approved):
LGTM!

finagolfin:
@shahmishal, easy review

.github/workflows/swift_package_test.yml
```diff
@@ -186,7 +186,7 @@ on:
       android_ndk_versions:
         type: string
         description: "The NDK versions to use when building with the Swift SDK for Android"
-        default: "[\"r27d\", \"r28c\"]"
+        default: "[\"r27d\", \"r28c\", \"r30\"]"
```
shahmishal: Should we drop 27?

finagolfin: Yeah, I have no problem dropping anything before 30, as these are only the defaults and devs can always specify what they want.

If you prefer some other default, just let me know.

shahmishal: Let's drop 27 and 28 from default

finagolfin: Done, I only kept the old ones around for the default because we still supported 6.3, built with NDK 27, when I submitted this pull adding NDK 30 last week. Now that 6.4 is out, built with NDK 30, and all prior NDKs are unsupported, this is the right default.

PR #316, opened 2026-09-11
Swift GitHub Actions Workflows Unification.

rnro:
Swift GitHub Actions Workflows Unification.

This PR aims to bring the best aspects of the swiftlang GitHub Actions workflows (this repository) and the SwiftNIO ([apple/swift-nio](https://github.com/apple/swift-nio/tree/main/.github/workflows)) workflows together. It does this chiefly through a new re-usable workflow which sits beside the existing workflow, and a hierarchy of supporting infrastructure.

Some key design features:

1) GitHub-release-based versioning
2) No skipped jobs
3) Scripts are cloned not curled
4) Custom matrix builds
5) Minimum Swift version detection

The unified design provides one workflow which can be adopted to run a range of common test and build configurations on a variety of platforms against multiple Swift versions. This workflow sits atop a "matrix generation" layer which takes inputs and produces a canonical work definition (in YAML or JSON). That definition is then expanded into one job per entry, each executing one slice of the work. Benchmarking is a further consumer of the same layer rather than part of the recommended test workflow.

The design is intended to be layered, so that adopters may use the whole stack, or customize portions beyond the level which the workflow allows. Adopters may:

* Use the `package_test.yml` workflow for the full suite of conveniences
* Use a custom workflow (perhaps with custom inputs) which calls `toolchain_matrix.yml` for the matrix and `execute_matrix.yml` to run its own command across it
* Generate the work definition in YAML or JSON by some other means, or hard-code it into a workflow, and pass it to `execute_matrix.yml`

The matrix generation layer is a Swift script, `generate-matrix.swift`, run by the matrix job on `ubuntu-latest`, which comes with the Swift, jq and yq it needs. It reads the workflow's inputs from the environment, applies the defaults and the minimum version filter, and writes the work definition to standard output. In SwiftNIO it was a bash script, which executes faster, but the unified version is much more complex, so Swift seemed like the more maintainable option. A test package under `tests/MatrixGeneratorValidator` covers the inputs, their defaults and failure modes.

As part of this work an effort was also made to reduce execution times, defaulting to running natively on the runner rather than inside a container where possible. The unified workflows install the toolchain with [swiftly](https://github.com/swiftlang/swiftly) on the runner rather than pulling a full Docker image. Containers remain available for a package that needs a particular distribution or system dependencies, either through `linux_use_docker` or by naming a distribution in `linux_os`. swiftly was previously used only for macOS toolchains, so using it on Linux is new to both upstreams.

## Migrating from this repository's existing workflow

The new workflow is currently available alongside existing infrastructure so a repository can control when they perform the migration.

When adopting the new workflow the most visible change is that the `*_exclude_swift_versions` inputs are gone, replaced with explicit supported Swift version lists and minimum-version detection which drops anything the package manifest cannot support. A repository that was only excluding old versions based on its Swift tools version can simplify its config.

Before:

```yaml
  tests:
    name: Test
    uses: swiftlang/github-workflows/.github/workflows/swift_package_test.yml@0.0.14
    with:
      enable_cross_pr_testing: true
      linux_exclude_swift_versions: "[{\"swift_version\": \"6.1\"}, {\"swift_version\": \"6.2\"}]"
      windows_exclude_swift_versions: "[{\"swift_version\": \"6.1\"}, {\"swift_version\": \"6.2\"}]"
```

After:

```yaml
  tests:
    name: Test
    uses: swiftlang/github-workflows/.github/workflows/package_test.yml@0.0.16
    with:
      enable_cross_pr_testing: true
```

The new architecture means that jobs which are not intended to be run will no longer appear as skipped:

Before:
Screenshot 2026-09-11 at 5 01 21 PM

After:
Screenshot 2026-09-11 at 5 27 20 PM

## Migrating from the SwiftNIO workflows

Changing from the SwiftNIO workflows means a change in spelling but no reduction in flexibility. Per-version inputs are now supplied in a map which reduces churn and synchronization problems when a new Swift version is released.

Before:

```yaml
  unit-tests:
    uses: apple/swift-nio/.github/workflows/unit_tests.yml@main
    with:
      linux_6_1_arguments_override: "-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error"
      linux_6_2_arguments_override: "-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error"
      linux_6_3_arguments_override: "-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error"
      linux_nightly_next_arguments_override: "-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error"
      linux_nightly_main_arguments_override: "--explicit-target-dependency-import-check error"
  benchmarks:
    uses: apple/swift-nio/.github/workflows/benchmarks.yml@main
  cxx-interop:
    uses: apple/swift-nio/.github/workflows/cxx_interop.yml@main
  construct-integration-test-matrix:   # hand-written: checkout, then build a matrix in a script
    runs-on: ubuntu-latest
    steps: [...]
  integration-tests:
    needs: construct-integration-test-matrix
    uses: apple/swift-nio/.github/workflows/swift_test_matrix.yml@main
```

After:

```yaml
  tests:                                     # unit tests and Cxx interop
    uses: swiftlang/github-workflows/.github/workflows/package_test.yml@0.0.16
    with:
      linux_version_overrides: |
        6.1: -Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error
        6.2: -Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error
        6.3: -Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error
        nightly-release: -Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error
        nightly-main: --explicit-target-dependency-import-check error
  benchmarks:
    uses: swiftlang/github-workflows/.github/workflows/benchmarks.yml@0.0.16
  integration-test-toolchains:               # replaces the hand-written matrix job
    uses: swiftlang/github-workflows/.github/workflows/toolchain_matrix.yml@0.0.16
  integration-tests:
    needs: integration-test-toolchains
    uses: swiftlang/github-workflows/.github/workflows/execute_matrix.yml@0.0.16
```

## Functionality new to this repository

The unification also brings some new functionality to this repository. Some inspired by SwiftNIO, some entirely new:

* **Simulator testing.** iOS was build-only; `xcodebuild` now builds *and* tests on simulators, and tvOS, watchOS and visionOS join it. Mac Catalyst is new to both upstreams.
* **Benchmarks across the matrix.**  `benchmarks.yml` measures performance with [Swift Package Benchmark](https://github.com/ordo-one/package-benchmark), checking committed thresholds under `Thresholds/<swift-version>/` on every platform and Swift version in the matrix, failing the job and printing a diff when one moves. This differs from the existing `performance_test.yml`, which compares a PR against its merge base in one container and posts the table as a PR comment. The existing workflow tells you what a change did to performance; the new one holds a package to a standard it has written down.
* **A C++ interoperability check.** A release-configuration build needs no dedicated input: it is a second labeled command, so `Release-Build Linux Swift 6.3` runs alongside `Debug-Test Linux Swift 6.3` from one call.
* **A Semantic Version PR label check** (`pull_request_label.yml`), which fails a PR that carries no SemVer label.
* **Importable Swift version matrix**, `toolchain_matrix.yml`, provides the current supported Swift version matrix for custom workflows to utilize. The output is plain YAML, so it can be filtered or extended with `yq` before it is executed.
* **Per-version overrides**, so one Swift version can take extra arguments, or a different command, without splitting the matrix by hand.

## Known limitations

* **A pinned tag still does not pin the scripts.** This is behavior inherited from the existing swiftlang workflows. `swift_package_test.yml` and `soundness.yml` check this repository out with no `ref:`, so they resolve to its default branch, and one variant hardcodes `ref: main`; this means a caller on a tag e.g. `@0.0.14` has always run that tag's YAML against the default branch's scripts. The unified workflows inherit that default. Resolving this should be possible but would require stamping versions as part of the release process.


.github/workflows/architecture.md
```diff
@@ -0,0 +1,47 @@
+# Swift Package Test Workflow Architecture
+
+## Aims
```
FranzBusch, line 3: Nit:
```suggestion
## Goals
```

.github/workflows/architecture.md
```diff
@@ -0,0 +1,47 @@
...
+
+The Swift Package Test workflow is intended to give maintainers of open-source Swift packages a low-friction way to give contributors confidence in the code they write. This means making testing against a range of Swift versions and platforms as simple as possible, and making the results as accessible as possible.
+
+This workflow builds on the GitHub Actions workflows from [apple/swift-nio](https://github.com/apple/swift-nio/tree/main/.github/workflows) and [swiftlang/github-workflows](https://github.com/swiftlang/github-workflows/tree/main/.github/workflows), taking from both repositories the key features listed below, which the unified architecture treats as requirements.
```
FranzBusch, line 7: I would drop this since it won't stand the test of time once NIO removes its workflows.

.github/workflows/execute_matrix.yml
```diff
@@ -0,0 +1,450 @@
...
+        default: "{}"
+      job_timeout:
+        type: number
+        description: "Timeout in minutes for each job (default: 60)"
+        default: 60
+      windows_job_timeout:
+        type: number
+        description: "Timeout in minutes for Windows jobs. 0 uses job_timeout. Windows installs a toolchain and Visual Studio Build Tools before building, so it often needs longer."
+        default: 0
+      freebsd_job_timeout:
```
FranzBusch, line 41: Why do we need different timeouts per platform? If we do should this be configured in the matrix instead?

.github/workflows/execute_matrix.yml
```diff
@@ -0,0 +1,450 @@
...
+                | .env = ($default_env + (.env // {}))
+             ]}')
+
+          missing=$(echo "$matrix" | jq -r '[.config[] | select((.command // "") == "") | .name] | join(", ")')
+          if [ -n "$missing" ]; then
+            echo "::error::No command for matrix entries: $missing. Set it on the entry or pass the workflow's command input."
+            exit 1
+          fi
+
+          # Reject unexpected platforms early.
```
FranzBusch, line 130:
```suggestion
          # Reject unexpected host platforms early.
```

.github/workflows/execute_matrix.yml
```diff
@@ -0,0 +1,450 @@
...
+      - name: Checkout the workflows repository
+        if: ${{ steps.workflows_source.outputs.needs_checkout == 'true' }}
+        uses: actions/checkout@v7
+        with:
+          repository: ${{ inputs.workflows_repository }}
+          ref: ${{ steps.workflows_source.outputs.ref }}
+          path: github-workflows
+          persist-credentials: false
+
+      - name: Check out linked PRs (macOS)
```
FranzBusch, line 215: Why does this step only run on macOS and the others don't need it? Can we at least add a comment on the why?

.github/workflows/package_test.yml
```diff
@@ -0,0 +1,510 @@
...
+        description: "Boolean to enable building with the Swift SDK for Android. Defaults to false."
+        default: false
+      enable_android_emulator_tests:
+        type: boolean
+        description: "Boolean to enable Android emulator testing after SDK builds. Defaults to false."
+        default: false
+      enable_cxx_interop:
+        type: boolean
+        description: "Boolean to enable Cxx interoperability check. Defaults to false."
+        default: false
```
FranzBusch, line 49: Why are all of these disabled by default? These are standard checks that we expect every swiftlang package to pass so IMO they should be enabled by default and require explicit opt-out.

.github/workflows/package_test.yml
```diff
@@ -0,0 +1,510 @@
...
+      cxx_interop_swift_versions:
+        type: string
+        description: "Swift version list (JSON array) for the Cxx interop check. Empty uses the newest release version in linux_swift_versions."
+        default: ""
+
+      # ----- Version lists (JSON arrays) -----
+      linux_swift_versions:
+        type: string
+        description: "Linux Swift version list (JSON array)."
+        default: '["6.1", "6.2", "6.3", "nightly-release", "nightly-main"]'
```
FranzBusch, line 59: Is this the only place that we need to update once a new Swift version is released?

.github/workflows/package_test.yml
```diff
@@ -0,0 +1,510 @@
...
+      linux_host_archs:
+        type: string
+        description: "Linux host architecture list (JSON array, e.g. '[\"x86_64\", \"aarch64\"]')."
+        default: '["x86_64"]'
+      macos_xcode_versions:
```
FranzBusch, line 64: Do we need this or can we just use the macOS_swift_versions instead?

.github/workflows/package_test.yml
```diff
@@ -0,0 +1,510 @@
...
+      swift_flags:
+        type: string
+        description: "Swift flags appended to release version builds."
+        default: ""
+      swift_nightly_flags:
+        type: string
+        description: "Swift flags appended to nightly version builds."
+        default: ""
+
+      # ----- Per-version overrides (JSON objects) -----
```
FranzBusch, line 225: Don't we need overrides for all of the platforms e.g. Android, WASM, etc.?

.github/workflows/package_test.yml
```diff
@@ -0,0 +1,510 @@
...
+        type: string
+        description: 'Docker --security-opt options for Linux containers (JSON/YAML array, e.g. ''["apparmor=unconfined"]'').'
+        default: "[]"
+
+      # ----- Minimum version detection -----
+      minimum_swift_version:
+        type: string
+        description: "Minimum Swift version. Empty auto-detects from Package.swift, 'none' disables filtering, or name a version explicitly."
+        default: ""
+      find_subdirectory_manifests:
```
FranzBusch, line 291: Why is this needed?

.github/workflows/package_test.yml
```diff
@@ -0,0 +1,510 @@
...
+        default: "tahoe"
+      macos_arch:
+        type: string
+        description: "macOS runner architecture label."
+        default: "ARM64"
+      macos_runner_pool:
+        type: string
+        description: "macOS self-hosted runner pool label."
+        default: "general"
+      macos_repository_owner:
```
FranzBusch, line 309: Isn't this the default on Github actions now that forks do not run actions by default?

.github/workflows/pull_request.yml
```diff
@@ -103,6 +103,151 @@ jobs:
       api_breakage_check_enabled: false
       license_header_check_project_name: "Swift.org"
 
+  matrix_generator_tests:
+    name: Matrix generator tests
+    # The runner image ships both Swift and yq, so nothing needs installing. It
+    # has to be mikefarah's yq, which is what the image carries: generate-matrix.swift
+    # invokes yq with no filter argument, and python-yq is a jq wrapper that
+    # requires one.
```
FranzBusch, line 111: Do we need this
```suggestion
```

.github/workflows/pull_request.yml
```diff
@@ -103,6 +103,151 @@ jobs:
...
+      linux_swift_versions: '["6.3"]'
+      # Runs the privilege-agnostic form the migration guide documents, so the
+      # advice is checked rather than only written down.
+      linux_setup_command: |
+        if command -v sudo >/dev/null 2>&1; then
+          sudo apt-get update -y -q && sudo apt-get install -y -q jq
+        else
+          apt-get update -y -q && apt-get install -y -q jq
+        fi
+        cd tests/TestPackage
```
FranzBusch, line 189: This feels somewhat unergonomic that we have to install various things and the also `cd`. Should we instead have a `package_path` input?

.github/workflows/pull_request_label.yml
FranzBusch, line 1: Could we separate the entire Semver check out into a separate PR? It is not tied to the matrix based builds at all.

.github/workflows/benchmarks.yml
```diff
@@ -0,0 +1,198 @@
...
+  contents: read
+
+on:
+  workflow_call:
+    inputs:
+      benchmark_package_path:
+        type: string
+        description: "Path to the directory containing the benchmarking package. Used only when benchmark_package_paths is empty."
+        default: "."
+      benchmark_package_paths:
```
FranzBusch, line 13: Do we need both or would just the plural be enough?

.github/workflows/scripts/matrix/generate-matrix.swift
```diff
@@ -0,0 +1,2313 @@
...
+import Foundation
+
+// MARK: - Configuration
+
+/// What the run was configured with, as the caller wrote it.
+struct Configuration {
+    /// The releases a platform tests by default, oldest first.
+    static let releases = ["6.1", "6.2", "6.3"]
+    /// The nightlies every platform runs: the next release's branch, and main.
+    static let nightlies = ["nightly-release", "nightly-main"]
```
FranzBusch, line 30: So there are defaults in the yml files and here. Do we need to update both? Do we need both to default? It would be great if we had just 1 place to update.

.github/workflows/scripts/matrix/generate-matrix.swift
FranzBusch, line 1: This file would benefit from splitting up into multiple files.

.github/workflows/benchmarks.yml
FranzBusch, line 1: This might be better added in a second PR since it is a net-new capability for this repo.

FranzBusch:
Overall this looks good. Thanks for working on it. I left some comments inline. One thing I would like to see added is the definition of the JSON/YML used by the matrix job so if somebody wants to hand-roll a matrix they have a reference to write it against.

> A pinned tag still does not pin the scripts. This is behavior inherited from the existing swiftlang workflows. swift_package_test.yml and soundness.yml check this repository out with no ref:, so they resolve to its default branch, and one variant hardcodes ref: main; this means a caller on a tag e.g. @0.0.14 has always run that tag's YAML against the default branch's scripts. The unified workflows inherit that default. Resolving this should be possible but would require stamping versions as part of the release process.

We should create an issue to track this since we ran into problems with this in the past.

.github/workflows/scripts/matrix/generate-matrix.swift
rnro: At the moment this is a script so I think it's limited to one file. We could make it a package but I think that has performance implications. If we can cache the built artifact then it can be a package, we can delegate work to dependencies, have a cleaner structure and we would have great performance. The execution time would then be under a second.

.github/workflows/package_test.yml
```diff
@@ -0,0 +1,510 @@
...
+        description: "Wasm Swift SDK version list (JSON array)."
+        default: '["6.3", "nightly-release", "nightly-main"]'
+      embedded_wasm_sdk_versions:
+        type: string
+        description: "Embedded Wasm Swift SDK version list (JSON array)."
+        default: '["6.3", "nightly-release", "nightly-main"]'
+      android_sdk_versions:
+        type: string
+        description: "Android Swift SDK version list (JSON array)."
+        default: '["6.3", "nightly-release", "nightly-main"]'
```
finagolfin, line 103: Better to make the first one "release" to signify the current release, then let people add 6.1 or whatever else they want also?

rnro: Hey @finagolfin, thanks for taking a look at the PR. I'm happy to be led by the consensus here, but I have discussed this with many Swift server repo admins and the (split) majority opinion was to only have sliding nightly versions. That way they can control when new Swift versions appear in their supported matrix. That said, I think nothing stops us from having both ways of spelling this so users could opt-in, but I think that's an enhancement.

